PR #16, opened 2018-01-30
Move docker includes into docker directory, add pom.xml.

cunningt:
Add a pom.xml to build with, move the docker file includes to a docker directory (required by fabric8-maven-plugin that the Dockerfile/docker includes not be in the root dir), and change gulp.js to output to docker/site.


astefanutti:
@cunningt thanks!

PR #28, opened 2018-04-21
Add licenses to docker image.

cunningt:
This is an attempt to add licenses to the hawtio-online docker image - I'm not sure this is the right way to go about it, but I figured I'd give it an initial shot.    This approach uses a few static files, but I'm not familiar enough with hawtio to know what else we need to list - do we need to list every node.js dependency?

This is a node license reporter from productization that could be used, but the package.json in hawtio-online doesn't contain dependencies, and a license entry also needs to be added : 

 https://docs.engineering.redhat.com/display/JPC/Product+Licensing+Information#ProductLicensingInformation-nodejs_license_reporterNode.jsLicenseReporter

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -22,6 +22,14 @@ RUN curl -sO http://nginx.org/keys/nginx_signing.key && \
     rm -f ./nginx_signing.key && \
     yum clean all
 
+# Copy licenses
+RUN mkdir -p /opt/fuse/licenses
+COPY licenses/licenses.css /opt/fuse/licenses
+COPY licenses/licenses.xml /opt/fuse/licenses
+COPY licenses/licenses.html /opt/fuse/licenses
+COPY licenses/mit.txt /opt/fuse/licenses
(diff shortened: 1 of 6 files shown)


astefanutti:
@cunningt thanks!

As this is a multi-modules project, the root `package.json` file doesn't have dependencies. However, it seems https://github.com/bucharest-gold/license-reporter does support that use case, by providing a _merge_ operation.

I can merge that PR and take it from there to try to automate the license files generation. Let me know if you have objections.

cunningt:
@astefanutti That sounds great to me.   Let me know if I can help at all.

astefanutti:
@cunningt I've been able to partially automate the licenses generation. This still requires to have bucharest-gold/license-reporter#304 fixed. In the meantime, I've updated the HTML report manually and created #29 to follow that up.

cunningt:
@astefanutti  That's great - I think manually for now is enough so that I can resolve https://issues.jboss.org/browse/ENTESB-6909

PR #38, opened 2019-07-08
OpenShift 4 support

astefanutti:


astefanutti:
Tested successfully on OpenShift 4.1. Remains documenting the creation of the client certificate and enabling caching of sub-requests.

astefanutti:
@jamesnetherton if you can give this a quick review that'd be appreciated.

jamesnetherton:
Looks good to me. 

One thing I wanted to understand before approving is the significance of the `update` verb on the `LocalSubjectAccessReview`. Why not `get`, `list` etc?

astefanutti:
> One thing I wanted to understand before approving is the significance of the update verb on the LocalSubjectAccessReview. Why not get, list etc?

For now, a user that can `update` the pod can call Jolokia. This is somehow equivalent to what was required in OpenShift 3 (was actually the `proxy` verb required). Currently, as the user can _mutate_ the app via Jolokia, it seems requiring `update` verb is more logical, as a default RBAC in the short term.

In the long term, this can be the place for a more advanced RBAC solution.

astefanutti:
@jamesnetherton thanks for the review!

astefanutti:
Remaining tasks:
* [ ] Detect OpenShift version client-side and use the master API proxy for OpenShift 3
* [ ] Document the creation of the client certificate secret
* [ ] Update images with Jolokia 1.6.2 and the default common name for Hawtio Online as authorised client of the Jolokia agent
* [ ] Cache sub-requests to the master API (authorisation and `podIP` resolution)
* [ ] Automate the generation of the client certificate secret in the operator
* [ ] OLM manifests

PR #39, opened 2019-08-06
fix: ENTESB-10521 Not visible items in Fuse Console Preferences

alexkieling:


packages/online/src/main.less
```diff
@@ -6,7 +6,6 @@
 @pf-padding: @grid-gutter-width / 2;
 
 #main {
```
astefanutti: Out of curiosity, where is the `#main` selector declared in the recent versions?

alexkieling:
I've tested the master and 1.x-redhat-7.x branches and they seem fine. I think this issue is only happening in versions up to 7.3.

packages/online/src/main.less
alexkieling: It's declared here: https://github.com/hawtio/hawtio-core/blob/4d91ea331156528df1342f12b935c4ca4e32de3d/src/app/page/page.component.ts#L89

astefanutti: Is this really matching the `#main` selector? I would have expected it should use a CSS attribute selector, like `[role=main]`.

astefanutti:
OK, that must explain why I could not reproduce on newer versions.

packages/online/src/main.less
alexkieling: Good point. I think recent versions don't set the "main" id anymore.

astefanutti: OK that was my understanding. Thanks for the confirmation.

PR #43, opened 2019-11-19
fix(osgi): ENTESB-12241 OSGi pages not working on OCP 4.1

akieling:
https://github.com/hawtio/hawtio-integration/pull/95

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -12,7 +12,7 @@
     "build": "gulp build"
   },
   "dependencies": {
-    "@hawtio/core": "~4.7.0",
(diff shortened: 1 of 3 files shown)


astefanutti:
Thanks. If you could create a PR for the master branch as well, that'd be great!

astefanutti:
@akieling ENTESB-12241 target release is set to 7.6. If that is confirmed, this should be merged into 1.x-redhat-7.x branch instead.

akieling:
Good catch. The target release was changed from 7.5 to 7.6 last week. I'll send a PR to 1.x-redhat-7.x.

PR #55, opened 2020-04-09
feat: Add RBAC support

astefanutti:
This PR adds RBAC to the management endpoint, i.e., the ability to control access to the Jolokia endpoints based on roles.

See https://issues.redhat.com/browse/ENTESB-8065 for more details.

TODO:
- [x] Support Jolokia bulk requests
- [x] Parameterize ACL file location, so that it can be secret mounted
- [x] Handle `canInvoke` operation, for seamless compatibility with client-side RBAC plugin
- [x] Migrate default ACL from Karaf into the new ACL format
- [x] Tests suite
- [x] Provide deployment templates

### DIFF
--- a/docker/ACL.yaml
+++ b/docker/ACL.yaml
@@ -1,6 +1,8 @@
 # TODO: complete and document
 
 default:
+  - list: admin, viewer
+  - read: admin, viewer
   - search: admin, viewer
   - /list.*/: admin, viewer
   - /get.*/: admin, viewer
--- a/docker/ACL.yaml
+++ b/docker/ACL.yaml
@@ -243,42 +243,42 @@ org.apache.karaf.bundle:
   resolve: admin
   restart: admin
   status: admin, viewer
-  setStartLevel(java.lang.String,int)[/([1-4])?[0-9]/,/.*/]: admin
+  /setStartLevel\(java\.lang\.String,int\)\[[1-4]?[0-9],.*\]/: admin
   setStartLevel: admin
-  start(java.lang.String)[/([1-4])?[0-9]/]: admin
+  /start\(java\.lang\.String\)\[[1-4]?[0-9]\]/: admin
   start: admin
-  stop(java.lang.String)[/([1-4])?[0-9]/]: admin
+  /stop\(java\.lang\.String\)\[[1-4]?[0-9]\]/: admin
   stop: admin
-  uninstall(java.lang.String)["0"]: #this is a comment, no roles can perform this operation
+  uninstall(java.lang.String)[0]: [] # no roles can perform this operation
   uninstall: admin
-  update(java.lang.String)[/([1-4])?[0-9]/]: admin
-  update(java.lang.String,java.lang.String)[/([1-4])?[0-9]/,/.*/]: admin
+  /update\(java\.lang\.String\)\[[1-4]?[0-9]\]/: admin
+  /update\(java\.lang\.String,java\.lang\.String\)\[[1-4]?[0-9],.*\]/: admin
   update: admin
 org.apache.karaf.config:
   listProperties: admin, viewer
-  appendProperty(java.lang.String,java.lang.String,java.lang.String)[/jmx[.]acl.*/,/.*/,/.*/]: admin
-  appendProperty(java.lang.String,java.lang.String,java.lang.String)[/org[.]apache[.]karaf[.]command[.]acl.+/,/.*/,/.*/]: admin
-  appendProperty(java.lang.String,java.lang.String,java.lang.String)[/org[.]apache[.]karaf[.]service[.]acl.+/,/.*/,/.*/]: admin
+  /appendProperty\(java\.lang\.String,java\.lang\.String,java\.lang\.String\)\[jmx\.acl\..*,.*,.*\]/: admin
+  /appendProperty\(java\.lang\.String,java\.lang\.String,java\.lang\.String\)\[org\.apache\.karaf\.command\.acl\..+,.*,.*\]/: admin
+  /appendProperty\(java\.lang\.String,java\.lang\.String,java\.lang\.String\)\[org\.apache\.karaf\.service\.acl\..+,.*,.*\]/: admin
   appendProperty(java.lang.String,java.lang.String,java.lang.String): admin
-  create(java.lang.String)[/jmx[.]acl.*/]: admin
-  create(java.lang.String)[/org[.]apache[.]karaf[.]command[.]acl.+/]: admin
-  create(java.lang.String)[/org[.]apache[.]karaf[.]service[.]acl.+/]: admin
+  /create\(java\.lang\.String\)\[jmx\.acl\..*\]/: admin
+  /create\(java\.lang\.String\)\[org\.apache\.karaf\.command\.acl\..+\]/: admin
+  /create\(java\.lang\.String\)\[org\.apache\.karaf\.service\.acl\..+\]/: admin
   create(java.lang.String): admin
-  delete(java.lang.String)[/jmx[.]acl.*/]: admin
-  delete(java.lang.String)[/org[.]apache[.]karaf[.]command[.]acl.+/]: admin
-  delete(java.lang.String)[/org[.]apache[.]karaf[.]service[.]acl.+/]: admin
+  /delete\(java\.lang\.String\)\[jmx\.acl\..*\]/: admin
+  /delete\(java\.lang\.String\)\[org\.apache\.karaf\.command\.acl\..+\]/: admin
+  /delete\(java\.lang\.String\)\[org\.apache\.karaf\.service\.acl\..+\]/: admin
   delete(java.lang.String): admin
-  deleteProperty(java.lang.String,java.lang.String)[/jmx[.]acl.*/,/.*/]: admin
-  deleteProperty(java.lang.String,java.lang.String)[/org[.]apache[.]karaf[.]command[.]acl.+/,/.*/]: admin
-  deleteProperty(java.lang.String,java.lang.String)[/org[.]apache[.]karaf[.]service[.]acl.+/,/.*/]: admin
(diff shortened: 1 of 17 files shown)


tadayosi:
@astefanutti Just nitpicking but is there any reason we need to keep using `var` for variable declarations?  My understanding is that `var` is error prone so it's better to use `let` or `const` whenever possible.

astefanutti:
@tadayosi unfortunately, the NGINX JavaScript module lags a bit behind w.r.t. ES6 compatibility. `let` and `const` variable declaration has yet to be supported and is tracked in https://github.com/nginx/njs/issues/105.

tadayosi:
@astefanutti Ah, ok. I think it's good to go then.

astefanutti:
@tadayosi, I think it's possible to implement the `canInvoke` calls directly in the Fuse console backend. The idea would be that it's possible to re-use the client-side RBAC plugin as-is. Do you think it'd be the best approach?

tadayosi:
@astefanutti It sounds like a good idea, but note that the client-side RBAC plugin is rather usability functionalities than security ones. You'd still need to protect Jolokia calls to JVM nodes from unauthorised users, otherwise an browser dev tool can easily tweak css to execute sensitive operations on the nodes.

astefanutti:
@tadayosi right, RBAC is still to be enforced at the backend side. If the `canInvoke` method is the only API used by the client-side RBAC plugin, then we would easily be able to have these _usability_ functionalities working as well, with a low effort.

tadayosi:
Cool. Yes, it's the only API but a collection of `canInvoke` methods:
https://github.com/hawtio/hawtio/blob/hawtio-2.10.0/hawtio-system/src/main/java/io/hawt/jmx/JMXSecurityMBean.java
So long as you can implement this mbean endpoint with the backend (node.js?), then the client-side RBAC plugin should work.

To be more precise, there is another extra API for the performance improvement, but for the time being we can focus on the basic feature and implement one by one.
https://github.com/hawtio/hawtio/blob/hawtio-2.10.0/hawtio-system/src/main/java/io/hawt/jmx/RBACRegistryMBean.java

FYI, this is where the plugin discovers the endpoint. I wonder how it can integrate with hawtio-online backend. Maybe we need to modify it as well.
https://github.com/hawtio/hawtio-integration/blob/master/plugins/rbac/ts/rbac.service.ts#L29

astefanutti:
Thanks for the details. I think it's possible to implement these in the NJS scripts.

That's a good point for the discovery mechanism. It has to be implemented as well, in addition to the `canInvoke` calls.

astefanutti:
@tadayosi @grgrzybek I still have to expose an option to configure the location of the ACL file. Apart from that, I've been able to test it successfully. Could you please give it a last round of review before I merge it?

tadayosi:
@astefanutti I'm reviewing it now. It may take a bit of while.

astefanutti:
@tadayosi thanks. No worries, I'm still working on the options for the ACL location and to turn on/off the RBAC feature. You may start by reviewing the ACL file `docker/ACL.yaml` which contains a bit of documentation, and that ressembles the Karaf ACL configuration files.

docker/ACL.yaml
```diff
@@ -0,0 +1,467 @@
...
+
+default:
+  - list: admin, viewer
+  - read: admin, viewer
+  - search: admin, viewer
+  - /list.*/: admin, viewer
+  - /get.*/: admin, viewer
+  - /is.*/: admin, viewer
+  - /set.*/: admin
+  - /.*/: admin
```
tadayosi, line 53: It's a bit non-intuitive to me that only the `default` requires a list but other keys use a map for the operation definitions. Is it a bad idea to use a map for both?

astefanutti: The current rules declared in the `default` keys intersect, mostly because of the `/.*/: admin` rule. So the precedence of other rules has to be guaranteed, and using a list instead of a map provides that guarantee. I haven't used a list elsewhere, as there should be no key with intersecting rules.

Using a map is less verbose, and probably has better direct lookup performance, but the rules must not intersect. Otherwise using a list is an option.

docker/nginx.js
```diff
@@ -12,52 +21,153 @@ function proxyJolokiaAgent(req) {
...
-  selfLocalSubjectAccessReview();
+  return selfLocalSubjectAccessReview('update')
+    .then(function (res) {
+      if (res.status !== 201) {
+        return Promise.reject(res);
+      }
+      var sar = JSON.parse(res.responseBody);
+      if (sar.allowed) {
+        // map the 'update' verb to the 'admin' role
+        return 'admin';
```
tadayosi, line 88: So it might have been described somewhere earlier already, but do we support only the fixed set of roles which is admin and viewer now, and will it remain so in the future as well?  It's not a big deal as I'm sure we've captured the customer's requirements being boiled down to the two roles; I just wanted to understand the impl correctly.

docker/test.js
```diff
@@ -0,0 +1,279 @@
+var fs = require('fs');
```
tadayosi, line 1: How can I run the test locally?  Is there any way to run the nginx script locally for test/dev purposes?  How do you test the scripts?

docker/rbac.js
```diff
@@ -0,0 +1,223 @@
...
+export default {
+  check: check,
+  intercept: intercept,
+  isCanInvokeRequest: isCanInvokeRequest,
+};
+
+var rbacMBean = 'hawtio:area=jmx,type=security';
+
+function isCanInvokeRequest(request) {
+  return request.type === 'exec' && request.mbean === rbacMBean && request.operation === 'canInvoke(java.lang.String)';
```
tadayosi, line 19: As far as I remember, the client-side Hawtio RBAC plugin uses the following three `canInvoke` methods. Is it ok here to check just one of them?
```js
  const MBEAN_ONLY = 'canInvoke(java.lang.String)';
  const OVERLOADED_METHOD = 'canInvoke(java.lang.String,java.lang.String)';
  const EXACT_METHOD = 'canInvoke(java.lang.String,java.lang.String,[Ljava.lang.String;)';
```

docker/ACL.yaml
tadayosi: OK, that makes sense. So only the `default` section requires a ordered list. Probably it would be good to have a line of comment that explains the reasoning why it requires a ordered list?

docker/nginx.js
tadayosi: The document _Fuse Console on OpenShift RBAC Support_ appears to explain it already:
> The idea of the proposed solution is to evolve step 3, from a unary RBAC mechanism, with only one assimilated admin, into a binary one, supporting two roles, viewer and admin, inferred from the permissions the user is granted on the Fuse application pod.

docker/rbac.js
tadayosi: Probably this `rbac.js` impl corresponds to `io.hawt.system.RBACMBeanInvoker` from the Hawtio Java backend which only uses `canInvoke(java.lang.String)`?

As for the design decision of `RBACMBeanInvoker`, I remember it comes from the restriction of Jolokia `Restrictor` API, which can only pass the operation name but no arg types/values.

I could always be wrong though :-)

tadayosi:
@astefanutti I left a few comments in the code but they are mostly due to my ignorance to the impl and are not requests for changes. Overall it looks good to me.

By now I came to realise that this solution is actually not a temporary workaround but rather should be the right way to ensure RBAC for Jolokia-enabled OpenShift apps. It's better than the standalone impl that only supports Karaf atm, because this solution is platform-agnositc and immediately supports all of them on OpenShift.

docker/ACL.yaml
astefanutti: Right, I've just added a comment in eba473e2bd1e511bc8f180d1921917cbc4268d09.

docker/nginx.js
astefanutti: I've also added some documentation in eba473e2bd1e511bc8f180d1921917cbc4268d09.

docker/rbac.js
astefanutti: From the tests I've been able to run so far, the `canInvoke(java.lang.String,java.lang.String)` and `canInvoke(java.lang.String,java.lang.String,[Ljava.lang.String;)` are not called by `hawtio-integration`. So that explains why I only implemented support for `canInvoke(java.lang.String)`.

So it seems my tests empirically confirms your analysis. Hopefully we are not both empirically and theoretically wrong ;)

docker/test.js
astefanutti: The tests must be run with `njs`, which is the CLI for the NGINX JS module: http://nginx.org/en/docs/njs/cli.html.

The CLI is available along with the NGINX JS module, that's installed in the Hawtio online container image, so I run the test from there, by running:

```
$ docker run --rm -it -v `pwd`:/test hawtio/hawtio-online bash
$ cd /test/docker
$ njs test.js
```

The `test.js` file imports the `nginx.js` file as an ES module. Unfortunately, until https://github.com/nginx/njs/issues/115 gets fixed, the main script cannot be an ES module. So I have to change `function proxyJolokiaAgent(req)` to `export default function proxyJolokiaAgent(req)` so that the test module can import the main script :(

I've mostly used the tests to support the development. It's clear running the tests should be streamlined, so that these can ultimately be used for continuous testing.

astefanutti: I've just found out that the latest release of NJS has introduced the `js_import` directive. So that should be possible to fix the issue. I'm trying right now... I'll keep you posted.

astefanutti: I've just pushed 5ccf2ea9206a0546ae48ce321659f404a6828ade so that no manual modifications is required anymore. You can just run the tests with `njs test.js`.

astefanutti:
I think this is ready. Thanks a lot @tadayosi for the thorough review!

Issue #59, opened 2020-07-10
Local run doesn't work with OpenShift 4.x


tadayosi:
In development, when running hawtio-online locally it shows pods that are running on OpenShift 4.x, but connecting to one of them it gets "Connection lost" errors in a new window and fails.
https://github.com/hawtio/hawtio-online#run

tadayosi:
Adding the following env vars to the app deployments makes external `curl` to the Jolokia endpoints work, but still local hawtio-online doesn't work.
```
AB_JOLOKIA_AUTH_OPENSHIFT=false
AB_JOLOKIA_PASSWORD_RANDOM=false
AB_JOLOKIA_OPTS=useSslClientAuthentication=false,protocol=https
```

astefanutti:
It's broken because of https://github.com/hawtio/hawtio-online/commit/752be3fe86df842a862a207aba28375534371f30.

For it to work, the local reverse proxy configuration must be updated, so that it rewrites the requests made to the `management` path:

https://github.com/hawtio/hawtio-online/blob/8c546f23f1ad6ccfe8fb09273778ad957581c7b0/gulpfile.js#L86

tadayosi:
Hi @astefanutti, fyi just reverting https://github.com/hawtio/hawtio-online/commit/752be3fe86df842a862a207aba28375534371f30 doesn't resolve the issue for me. There appears to be other places that still invokes `/management/namespaces/...` paths.

astefanutti:
@tadayosi ah right, there is also https://github.com/hawtio/hawtio-online/blob/8c546f23f1ad6ccfe8fb09273778ad957581c7b0/packages/online/src/discover/discover.module.ts#L77
Better fixing the reverse proxying anyway 😉.

PR #65, opened 2020-08-21
fix(dev): Local run doesn't work with OpenShift 4.x #59

tadayosi:
Attempt to fix the local run issue.

@astefanutti Pls review this to quickly check if I'm going the right path.


tadayosi:
@astefanutti Thanks. Simply documenting it looks acceptable and sufficient for now to me, too.

astefanutti:
Thanks @tadayosi, that's a great improvement!

Issue #66, opened 2020-08-26
Support different Jolokia ports & paths


tadayosi:
If I'm not mistaken, currently hawtio-online hardcodes the Jolokia port to `8778` and assumes it's available at `/jolokia`. However, for instance Artemis broker exposes the Jolokia endpoint with the following port and the path `/console/jolokia` by default:
```yaml
      ports:
        - containerPort: 8161
          name: console-jolokia
          protocol: TCP
```

The main focus here is to recognise Artemis brokers on OpenShift but it would be also flexible that hawtio-online can pick up arbitrary port numbers and Jolokia paths given they are specified in the application deployment spec.

tadayosi:
@astefanutti Is there a good way to specify the Jolokia path in an application deployment spec?  I don't think `spec.containers[].ports[]` accepts key like `path`. Should we introduce an env variable like `HAWTIO_JOLOKIA_PATH` as follows?
```yaml
spec:
  containers:
    - env:
        - name: HAWTIO_JOLOKIA_PATH
          value: /console/jolokia
```

astefanutti:
@tadayosi right, while the port can be retrieve from the container port, the path would have to be provided elsewhere in the pod metadata. I would think an annotation would be the best place, e.g.:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: artemis-broker
  annotations:
   "hawt.io/jolokiaPath": /console/jolokia
```

tadayosi:
Annotations seem much better!  Thanks for your suggestion.

PR #69, opened 2020-09-04
DO NOT MERGE - feat(nginx): support basic auth for connection between nginx proxy and Jolokia endpoint

tadayosi:
This is still work-in-progress and not a complete fix, so please don't merge it.

I still haven't abandoned the approach to make it possible that the nginx proxy connects to a Jolokia endpoint provided by an Artemis broker with basic auth, instead of client certificate auth. This is preferable in that you don't need to nudge Artemis dev team to make the Artemis container image support client certificate auth and we can complete the task within hawtio-online. That's why I still stick to the option :-)

I've come up with a _hacky_ way to support basic auth. The patch is still incomplete but has reached the point where at least it partially works, so I'd like to share it with you and ask for your opinions.

@astefanutti What do you think about this approach?  Could you please review it and let me know your thoughts?

One concern is the way it passes the auth credentials to the nginx proxy server, which currently just embeds it to the proxied url. Maybe we can pass it in a request header with name something like `X-Hawtio-Authorization`.

### DIFF
--- a/docker/nginx-gateway.conf
+++ b/docker/nginx-gateway.conf
@@ -1,5 +1,7 @@
 js_import gateway from conf.d/nginx.js;
 
+js_set $credentials gateway.readCredentials;
+
 proxy_cache_path /var/cache/nginx/pods levels=1:2 keys_zone=pods:1m max_size=2m inactive=60m use_temp_path=off;
 proxy_cache_path /var/cache/nginx/rbac levels=1:2 keys_zone=rbac:100k max_size=1m inactive=60m use_temp_path=off;
 proxy_cache_path /var/cache/nginx/rbac2 levels=1:2 keys_zone=rbac2:100k max_size=1m inactive=60m use_temp_path=off;
@@ -122,8 +124,8 @@ server {
         proxy_ssl_certificate     /etc/tls/private/proxying/tls.crt;
         proxy_ssl_certificate_key /etc/tls/private/proxying/tls.key;
         proxy_ssl_session_reuse   on;
-        # Do not foward authorization header
-        proxy_set_header          Authorization "";
+        # Send basic auth credentials (for Artemis) or otherwise empty
+        proxy_set_header          Authorization $credentials;
     }
 
     # Get the OpenShift console URL
--- a/docker/nginx.js
+++ b/docker/nginx.js
@@ -6,7 +6,7 @@ import RBAC from 'rbac.js';
 
 var isRbacEnabled = typeof process.env['HAWTIO_ONLINE_RBAC_ACL'] !== 'undefined';
 
-export default { proxyJolokiaAgent };
+export default { proxyJolokiaAgent, readCredentials };
 
 // Only Jolokia requests using the POST method are currently supported,
 // as this is more comprehensive and it's what the front-end uses.
@@ -22,6 +22,14 @@ function proxyJolokiaAgent(req) {
   var pod = parts[3];
   var port = parts[4];
   var path = parts[5];
+
+  // extract basic auth credentials
+  var credentials;
+  if (pod.includes('@')) {
+    var split = pod.split('@');
+    credentials = split[0];
+    pod = split[1];
+  }
 
   function response(res) {
     for (var header in res.headersOut) {
@@ -75,7 +83,11 @@ function proxyJolokiaAgent(req) {
   }
 
   function callJolokiaAgent(podIP, request) {
-    return req.subrequest(`/proxy/${protocol}:${podIP}:${port}/${path}`, { method: req.method, body: request });
+    var args = '';
+    if (credentials) {
+      args = `credentials=${credentials}`;
+    }
+    return req.subrequest(`/proxy/${protocol}:${podIP}:${port}/${path}`, { method: req.method, body: request, args: args });
   }
 
   function proxyJolokiaAgentWithoutRbac() {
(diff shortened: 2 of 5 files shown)


astefanutti:
Thanks @tadayosi!

Here are my open points related to the basic authentication approach:

* From a UX standpoint: how are the credentials provided? Is the user prompted for each brokers?
* From a security standpoint: my understanding is that the credentials have to be forwarded to the Jolokia endpoints. While this is acceptable for the embedded console, as other mechanisms exist to guarantee the credentials are not stolen, it may be an issue in the case of a centralised console that automatically discovers endpoints. Indeed, a rogue pod could trick the discovery mechanism and make the centralised console call a fake Jolokia endpoint hosted by that pod, that would receive the credentials.

tadayosi:
Thanks @astefanutti!  They are valid points, and especially security is my primary concern too.

> - From a UX standpoint: how are the credentials provided? Is the user prompted for each brokers?

Yes, the user is prompted for each broker like the following:

Screenshot from 2020-09-07 13-39-39

I'm optimistic about the UX standpoint though, as the behaviour is exactly same as the standalone Connect tab where the user is trying to connect to a remote node which is protected with basic auth. So the user won't be surprised.

> - From a security standpoint: my understanding is that the credentials have to be forwarded to the Jolokia endpoints. While this is acceptable for the embedded console, as other mechanisms exist to guarantee the credentials are not stolen, it may be an issue in the case of a centralised console that automatically discovers endpoints. Indeed, a rogue pod could trick the discovery mechanism and make the centralised console call a fake Jolokia endpoint hosted by that pod, that would receive the credentials.

This is very important. So if I'm not mistaken, generally speaking we should not rely on any authn strategies other than the client certificate auth for any proxy-based inter-node communications, including those via [k8s services](https://kubernetes.io/docs/concepts/services-networking/service/). It is because otherwise a malicious pod can always abuse the pod selector labels to get some kind of credentials. Is that correct?

If so, I'm convinced that we should indeed abandon the basic auth approach for hawtio-online and confidently ask the Artemis team to apply the jolokia agent jar to their Artemis containers :-)

tadayosi:
In addition, it's not still clear to me why the client certificate auth can be said safer in this case. Is it because there are less chances of the certificates being stolen, and even if they are stolen they are not identifications on a specific system on the container?

astefanutti:
> In addition, it's not still clear to me why the client certificate auth can be said safer in this case. Is it because there are less chances of the certificates being stolen, and even if they are stolen they are not identifications on a specific system on the container?

During TLS client certificate authentication, only the public part of the client certificate is sent to the server, which cannot be used to impersonate the client. The client private key remains on the client side. I've found https://stackoverflow.com/questions/52879121/how-does-sending-client-certificate-not-expose-the-client-to-impersonation, but I guess it is possible to find other sources.

Besides, this is the identity of the Hawtio console, not that of the end-users.

astefanutti:
> This is very important. So if I'm not mistaken, generally speaking we should not rely on any authn strategies other than the client certificate auth for any proxy-based inter-node communications, including those via k8s services. It is because otherwise a malicious pod can always abuse the pod selector labels to get some kind of credentials. Is that correct?

Right.

> If so, I'm convinced that we should indeed abandon the basic auth approach for hawtio-online and confidently ask the Artemis team to apply the jolokia agent jar to their Artemis containers :-)

This is my conviction too. I think we should do a PoC to evaluate how complex that is to enable client certificate authentication in Artemis brokers, so we can iterate with the Artemis team. It seems possible from the documentation: https://activemq.apache.org/components/artemis/documentation/2.8.0/security.html. Alternatively, using the Jolokia agent would be an option. This is what has been done in Fuse quickstarts. Default CN are configured to be authorised, and the [service signing certificate authority](https://docs.openshift.com/container-platform/4.5/security/certificates/service-serving-certificate.html), that is already mounted in every pod in OpenShift, is used for validation.

Last but not least, it would enable a consistent UX, alleviating the need for users to be prompted for credentials.

tadayosi:
Thanks for those answers!  It's convincing enough to withdraw this branch. I'll leave the idea at the different branch [basic-auth-support](https://github.com/hawtio/hawtio-online/tree/basic-auth-support) just for future references.

tadayosi:
Just for our records,

>  I think we should do a PoC to evaluate how complex that is to enable client certificate authentication in Artemis brokers, so we can iterate with the Artemis team. It seems possible from the documentation: https://activemq.apache.org/components/artemis/documentation/2.8.0/security.html. Alternatively, using the Jolokia agent would be an option.

I priotise the latter option over the former option, because there is a use case [1] in AMQ that allows users to monitor directly the broker via the console jolokia endpoint. While it should still be possible, requiring to turn on client certification auth by default for the entire Artemis console for connectivity to hawtio-online seems a bit too demanding for casual monitoring use cases like that. We can explore the former option later and it would be doable because #66 is already in place.

I've done a small PoC [2] for the latter option and confirmed it indeed enables Artemis to be detectable from hawtio-online.

[1] https://access.redhat.com/documentation/en-us/red_hat_amq/7.7/html/deploying_amq_broker_on_openshift/assembly_br-broker-monitoring_broker-ocp#proc_br-monitoring-broker_broker-ocp
[2] https://github.com/tadayosi/artemis-jolokia-docker

astefanutti:
@tadayosi awesome! I agree the Jolokia agent approach seems less involved. I think that first version will demonstrate the benefit of a centralized management console, and it'll be easier to iterate with the Artemis team, so that we can eventually get rid of the standalone console and the basic authentication, and solely rely on the same authentication and authorisation mechanisms.

Issue #80, opened 2021-03-17
Move Docker image to UBI RHEL 8


tadayosi:
For better security and stability.

tadayosi:
Done with 0f04bb9e4ba48b54b63a85fec1d0188fe3c7f0fa

PR #81, opened 2021-03-18
feat(docker): move Docker image to UBI RHEL 8 #80

tadayosi:
@cunningt Could you have a look at this?  Especially on whether the usage of the nginx versions is ok from productisation point of view. I was hinted that the previous usage of the mainline 1.17 version (instead of this stable) might have been due to productisation issues.
```
ENV NGINX_VERSION 1.18.0-1.el8
ENV NGINX_MODULE_NJS_VERSION 1.18.0.0.4.3-1.el8
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/node:12.17.0 as builder
+FROM docker.io/node:12.21.0 as builder
 
 RUN yarn global add gulp-cli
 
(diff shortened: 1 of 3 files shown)


cunningt (approved):
Looks good to me!

tadayosi:
@djcoleman Just in case, could you have a look at this change, too, please?  Do you think it's ok from the productisation point of view?  Thank you!

tadayosi:
According to @djcoleman this repo is not used for docker image productisation, so it's safe to go ahead and merge  it.

Issue #23, opened 2018-02-20
Support Kubernetes


astefanutti:

ninckblokje:
Hi,

Is it possible to run hawtio-online on Kubernetes, not on OpenShift?

astefanutti:
For the time being it is not possible, mostly due to the way authentication is setup. I plan to work on bringing Kubernetes support ASAP.

ninckblokje:
Ok. Thanks for the quick reply!

mxandeco:
Hey, is there any progress on this? is the authentication the only blocker for the integration?

astefanutti:
It has yet to be done. The main area task is to enable token-based authentication. The UI should be slightly adapted to managed the lack of user information (as provided by the OpenShift API). On the backend-side, the bits relying on the service signing certificate should be adapted as well.

mariuszkopowski:
Hi any news on that ?

anerathil:
Hey, is there any progress on this?

ilyesAj:
any updates here ?

tadayosi:
Fixed. The remaining brushing-up will be done at #89

PR #88, opened 2021-12-07
Support Kubernetes

tadayosi:
#23

~Still work in progress. Do not merge.~
Initial version done.


oscerd (approved):

tadayosi:
Now the Online console pod discovery works. Next I'll fix the Integration console.

Screenshot from 2021-12-09 15-46-09

tadayosi:
Initial version done.

You can do `yarn deploy:k8s` to deploy a hawtio-online instance to k8s cluster.

Right now, a Jolokia app needs to disable SSL client auth (i.e. `protocol=https,useSslClientAuthentication=false`) to get it connected because service-ca is not available for k8s.

@astefanutti Please have a look if you're not on holiday yet.

tadayosi:
I forgot to add one more step to make it really work on a pure k8s cluster. Currently the `nginx-gateway.conf` still depends on `hawtio-online-tls-proxying` secret which should be removed. I'll push one more commit soon.

I intend to make the form-based auth work on both OpenShift and k8s. At this moment, you can try it on OpenShift.

astefanutti:
Woot!

A Camel K integration running on Minikube and managed by Hawtio:

Screenshot 2021-12-16 at 11 04 42
Screenshot 2021-12-16 at 11 04 51

packages/common/src/openshift/openshift.service.ts
```diff
@@ -35,8 +35,8 @@ namespace Online {
       if (this.is(HawtioMode.Cluster)) {
         const projects_client = this.K8SClientFactory.create(
           {
-           kind: KubernetesAPI.WatchTypes.PROJECTS,
-           labelSelector: _.get(configManager.config, "online.projectSelector", null),
+            kind: KubernetesAPI.WatchTypes.PROJECTS,
```
astefanutti, line 38: This fails on vanilla Kubernetes as this translates to querying an OpenShift API. This must be changed to `WatchTypes.NAMESPACES` dynamically.

packages/common/src/management/management.helper.ts
```diff
@@ -1,11 +1,28 @@
...
-    const protocol = pod.metadata.annotations['hawt.io/protocol'] || 'https';
-    const jolokiaPath = pod.metadata.annotations['hawt.io/jolokiaPath'] || '/jolokia';
+    const protocol = getAnnotation(pod, 'hawt.io/protocol', 'https');
+    const jolokiaPath = getAnnotation(pod, 'hawt.io/jolokiaPath', '/jolokia');
     return `/management/namespaces/${namespace}/pods/${protocol}:${name}:${port}${jolokiaPath}`;
   }
 
+  function getAnnotation(pod: Pod, name: string, defaultValue: string): string {
+    if (pod.metadata.annotations) {
+      return pod.metadata.annotations[name];
```
astefanutti, line 23: It should return `defaultValue` in case the annotation value is undefined.

deploy/bases/service.yml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: hawtio-online
+  labels:
+    app: hawtio
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: hawtio-online-tls-serving
```
astefanutti, line 8: I think this should be moved to an OpenShift "overlay".

astefanutti (approved):
That's awesome!

The major area I see that remains is smoothing the user experience, in terms of configuration and documentation.

For the configuration, I can think of:
* Having a Kustomize configuration for vanilla Kubernetes
* Have an extra overlay for OpenShift (and maybe removing the existing resources at the project root)
* Decide how to handle HTTPS serving or not:
    * We can go for documenting manual generation of the serving certificate
    * Or for relying on https://cert-manager.io/ 

For the documentation, I can think of:
* Updating the README to reflect that PR work
* Adding a section on how to create a ServiceAccount and retrieve its token to be used in the form login

deploy/bases/configmap-hawtio-online.yml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: hawtio-online
+data:
+  hawtconfig.json: |
+    {
+      "about": {
+        "title": "Hawtio OpenShift Console",
```
astefanutti, line 9: I'd suggest to remove "OpenShift" occurrences.

packages/common/src/management/management.helper.ts
tadayosi: Thanks. Fixed.

tadayosi:
Thanks for your feedback!  I'll tackle the remaining items from your feedback in a separate issue.

astefanutti:
Thank you for this awesome work!

Issue #89, opened 2021-12-17
Improve UX for Kubernetes support


tadayosi:
Follow-up on #23.

From https://github.com/hawtio/hawtio-online/pull/88#pullrequestreview-833883742:
> The major area I see that remains is smoothing the user experience, in terms of configuration and documentation.
>
> For the configuration, I can think of:
> - Having a Kustomize configuration for vanilla Kubernetes
> - Having an extra overlay for OpenShift (and maybe removing the existing resources at the project root)
> - Deciding how to handle HTTPS serving or not:
>   - We can go for documenting manual generation of the serving certificate
>   - Or for relying on https://cert-manager.io/
> 
> For the documentation, I can think of:
> - Updating the README to reflect that PR work
> - Adding a section on how to create a ServiceAccount and retrieve its token to be used in the form login

**Already done:**
- [x] Having a Kustomize configuration for vanilla Kubernetes
- [x] Having an extra overlay for OpenShift
 
**To do:**
- [x] Removing the existing resources at the project root
- [x] Deciding how to handle HTTPS serving or not
- [x] Updating the README to reflect that PR work
- [x] Adding a section on how to create a ServiceAccount and retrieve its token to be used in the form login

PR #95, opened 2022-01-26
doc(k8s): reorg README and update with the new deployment method #89

tadayosi:
- How to create a ServiceAccount for k8s deployment will be added later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,66 +1,124 @@
 # Hawtio Online
 
-An Hawtio console that eases the discovery and management of _hawtio-enabled_ <sup>[1](#f1)</sup> applications deployed on OpenShift.
+[![Test](https://github.com/hawtio/hawtio-online/actions/workflows/test.yml/badge.svg)](https://github.com/hawtio/hawtio-online/actions/workflows/test.yml)
+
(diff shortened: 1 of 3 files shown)


oscerd (approved):

README.md
```diff
@@ -1,220 +1,174 @@
...
-  dc/hawtio-online deploys istag/hawtio-online:latest 
-    deployment #1 deployed 2 minutes ago - 1 pod
-```
+### Kubernetes
 
-Open the route URL displayed above from your Web browser to access the Hawtio Online console.
+#### Proxying certificates
 
-## OpenShift 4
+For Kubernetes, proxing certificates are not supported, so you don't need to go through the steps.
```
astefanutti, line 48: Should we rather say it's disabled by default. It's possible to use a proxying client certificate, even for Kubernetes, the may difference with OpenShift is that a custom CA must be generated / used, and mounted / configured into the Jolokia agent.

astefanutti (approved):
Great!

README.md
tadayosi: Thanks. Yes, you are right. I'll change that later.

astefanutti: Agreed, that can be improvement in a subsequent PR.

PR #96, opened 2022-01-27
doc(k8s): add doc on creating serviceaccount for form auth #89

tadayosi:
The final part of #89


README.md
```diff
@@ -168,6 +168,21 @@ otherwise:
...
+## Authentication
+
+Hawtio Online currently supports two authentication modes: `oauth` and `form`, which is configured through `HAWTIO_ONLINE_AUTH` environment variable on Deployment.
+
+| Mode | Description |
+| ---- | ----------- |
+| oauth | Authenticates through [OpenShift OAuth server](https://docs.openshift.com/container-platform/4.9/authentication/understanding-authentication.html). It is available only on OpenShift. |
+| form | Authenticates with a [ServiceAccount](#creating-user-for-form-authentication) throught the Hawtio login form. |
+
+### Creating user for Form authentication
```
astefanutti, line 180: It may be more correct to say that using a token from a ServiceAccount is one option, as there are other mechanisms that could be used like: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-token-file. So I would suggest, to emphasise that this is one example, but any bearer token can be used.

docs/create-user.md
```diff
@@ -0,0 +1,52 @@
...
+```sh
+$ cat <<EOF | kubectl apply -f -
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: hawtio-user
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
```
astefanutti, line 33: Maybe it can be stressed that this is an example, with broad permissions that should be refined carefully.

astefanutti (approved):
That's great!

README.md
tadayosi: Thanks!  Updated.

docs/create-user.md
tadayosi: Thanks!  Updated.

PR #142, opened 2023-07-14
chore(build): update yarn to 3.6.1 and hold yarn in repo

tadayosi:
My first contribution to 2.x. @phantomjinx Please have a look.

* Upgrade to Yarn 3.6.1 (latest stable).
* Copy yarn 3.6.1 and plugin under .yarn/ as otherwise builds cannot be portable among developers.

`*/` in .gitignore is intentionally deleted. As far as I remember, it was added to avoid copying unnecessary files to the docker build image but now we have .dockerignore and I'm not sure if this entry was really needed. Let's revisit it when we are to build a docker image for hawtio-online v2.


phantomjinx (approved):
LGTM

Issue #152, opened 2023-10-02
The use of jolokia GET requests with body is to be discontinued


phantomjinx:
The jolokia [documentation](https://jolokia.org/reference/html/clients.html#client-javascript) outlines the use of HTTP requests being made using the GET or POST methods with a body.

Despite this, the nginx.js [script](https://github.com/hawtio/hawtio-online/blob/main/docker/nginx.js#L177-L186) has not supported GET requests created in this fashion. Therefore, it will not be implemented in version 2.x either. The result will be that all requests using the GET method must utilise the URI path while POST methods will continue to use the body.

For example, all tests in the [njs test script](https://github.com/hawtio/hawtio-online/blob/main/docker/test.js#L37) now have a POST method added to each request.

tadayosi:
@phantomjinx I still don't get what's the issue here.

> The jolokia [documentation](https://jolokia.org/reference/html/clients.html#client-javascript) outlines the use of HTTP requests being made using the GET or POST methods with a body.
> 
> Despite this, the nginx.js [script](https://github.com/hawtio/hawtio-online/blob/main/docker/nginx.js#L177-L186) has not supported GET requests created in this fashion.

Do you mean `nginx.js` should have supported GET requests **with** a body?  Then I don't think Jolokia has ever supported GET with a body, either. Please help me to understand what's really the problem with it.

phantomjinx:
For the `njs` tests to ever have passed, the `nginx.js` script would have allowed GET / Body requests.

The `nginx.js` script was changed (???) at some point to not [support](https://github.com/hawtio/hawtio-online/blob/main/docker/nginx.js#L182-L227) this sittuation, if it ever did??

So having witnessed the `njs` tests all returning 502 errors, I went to check the jolokia documentation and found that, in limited circumstances, such GET / Body requests are legitimate.

This issue though is confirming that going forward, with the use-case being such a corner issue, we are not going to change anything to support such requests. In summary, I am confirming the status-quo and updated the `njs` tests accordingly.

tadayosi:
Btw, when I run the njs tests on 1.x branch it passes all fine:
```console
$  yarn test:nginx 
yarn run v1.22.19
$ cd docker/ && HAWTIO_ONLINE_RBAC_ACL= njs test.js
Done in 0.09s.
```
Could you confirm it's the same with you?  If not, something should be different in your specific environment.

phantomjinx:
[...]

PR #154, opened 2023-10-11
fix: Updates njs tests to return a failure

phantomjinx:
- Executing tests on github, even if they fail the test suite has a 0 exit code so everything is green.
 
- Updates the test script to throw an error if there is a failure and the script exit's 1 instead
 
- Pass into the report function the expected result status code to compare with the actual which then determines if there is a failure.

See #152 for discussion


tadayosi (approved):
thanks

PR #178, opened 2023-11-02
Pick latest changes from 2.x branch

phantomjinx:
- list-projects scope permission for cluster install config
- Corrects install yarn executions to deal with new patches directory

### DIFF
--- a/docker/osconsole/config.sh
+++ b/docker/osconsole/config.sh
@@ -12,7 +12,7 @@ openshift_config_cluster() {
   "openshift": {
     "oauth_metadata_uri": "/master/.well-known/oauth-authorization-server",
     "oauth_client_id": "${HAWTIO_OAUTH_CLIENT_ID:-hawtio-online}",
-    "scope": "user:info user:check-access role:edit:*",
+    "scope": "user:info user:check-access user:list-projects role:edit:*",
     "web_console_url": "${OPENSHIFT_WEB_CONSOLE_URL:-}",
     "cluster_version": "${OPENSHIFT_CLUSTER_VERSION:-}"
   }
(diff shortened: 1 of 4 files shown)


oauthclient.yml
```diff
@@ -1,7 +1,7 @@
-apiVersion: v1
+apiVersion: oauth.openshift.io/v1
 kind: OAuthClient
 metadata:
-  name: hawtio-online-dev
+  name: hawtio-online
```
tadayosi, line 4: I think this was intentional as otherwise it should overwrite what the normal install process installs as the oauth client. The one for local dev should be kept as a different client.

docker/osconsole/config.sh
```diff
@@ -12,7 +12,7 @@ openshift_config_cluster() {
   "openshift": {
     "oauth_metadata_uri": "/master/.well-known/oauth-authorization-server",
     "oauth_client_id": "${HAWTIO_OAUTH_CLIENT_ID:-hawtio-online}",
-    "scope": "user:info user:check-access role:edit:*",
+    "scope": "user:info user:check-access user:list-projects role:edit:*",
```
tadayosi, line 15: I wonder why this starts to be needed for hawtio-online v2. What changes require it?

phantomjinx: It probably was always supposed to be there but I took it out during the migration, perhaps. Looking back at [HO-1.0](https://github.com/hawtio/hawtio-online/blob/1.x-redhat-7.x/docker/osconsole/config.sh#L15) was the path to the solution.

oauthclient.yml
phantomjinx: Backed out. Good catch. Thx.

tadayosi (approved):

PR #196, opened 2023-11-14
HAWNG-6: Adds custom 404 and 50x error pages

phantomjinx:
* Avoids revealing type or version of web server used by displaying custom error pages


tadayosi (approved):
Thanks!

PR #198, opened 2023-11-17
Upgrade node 18

phantomjinx:
- Supercedes https://github.com/hawtio/hawtio-online/pull/190 and ensures that code compiles correctly by changing the monkey patching of window.fetch to a fetch intercepter.

- Supercedes #197 by breaking down into requisite concerns.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
   "packageManager": "yarn@3.6.1",
   "engines": {
     "yarn": ">=3.6.0",
-    "node": ">=16"
+    "node": ">=18"
   },
   "dependencies": {
     "jasmine": "^5.1.0"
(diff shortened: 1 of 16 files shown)


packages/management-api-app/src/management.tsx
```diff
@@ -32,7 +32,7 @@ export const Management: React.FunctionComponent = () => {
   const [error, setError] = useState<Error | null>()
   const [username, setUsername] = useState('')
 
-  const [pods, setPods] = useState<ManagedPod[]>([])
+  const [pods, setPods] = useState<ManagedPod[] | null>(null)
```
tadayosi, line 35: Like I said, instead of introducing adhoc `null` type, we should use `isLoading` flag to decide whether it's a valid state of `[]` or just still loading.

tadayosi: But I can't see how it's related to Node upgrade.

phantomjinx: Sneaked into this commit as a result of the fix for the monkey patching. Now removed.

PR #199, opened 2023-11-17
Upgrades hawtio/react to 0.9.2

phantomjinx:
* Supercedes https://github.com/hawtio/hawtio-online/pull/192
* As a result, of this upgrade, provides fix for test app login forms
* Supercedes https://github.com/hawtio/hawtio-online/pull/197 by breaking down into requisite concerns.


phantomjinx:
This PR will need to have #198 be merged then it will need to be rebased in order for the tests to pass. The fix for resolving webpack node_modules directories is required from #198.

packages/kubernetes-api-app/src/login/AuthLoginPage.tsx
```diff
@@ -1,27 +1,40 @@
...
+  const { isLogin, userLoaded } = useUser()
   const { hawtconfig, hawtconfigLoaded } = useHawtconfig()
   const { plugins, pluginsLoaded } = usePlugins()
 
-  if (!hawtconfigLoaded && !pluginsLoaded) {
+  useEffect(() => {
+    if (isLogin) {
+      navigate('/')
+    }
+  }, [isLogin, navigate])
```
tadayosi, line 20: Looks like you separated the pull reqs but haven't looked at my previous reviews at all?

I said:
While I know hawtio/react does the same code, now I wonder why we need `useEffect`. (Sorry it must be me who first introduced it...)  Can't we just put it in the main body of the component? :thinking:

packages/management-api-app/src/login/AuthLoginPage.tsx
```diff
@@ -1,27 +1,40 @@
...
+  const { isLogin, userLoaded } = useUser()
   const { hawtconfig, hawtconfigLoaded } = useHawtconfig()
   const { plugins, pluginsLoaded } = usePlugins()
 
-  if (!hawtconfigLoaded && !pluginsLoaded) {
+  useEffect(() => {
+    if (isLogin) {
+      navigate('/')
+    }
+  }, [isLogin, navigate])
```
tadayosi, line 20: same as above.

packages/oauth-app/src/login/AuthLoginPage.tsx
```diff
@@ -1,27 +1,40 @@
...
+  const { isLogin, userLoaded } = useUser()
   const { hawtconfig, hawtconfigLoaded } = useHawtconfig()
   const { plugins, pluginsLoaded } = usePlugins()
 
-  if (!hawtconfigLoaded && !pluginsLoaded) {
+  useEffect(() => {
+    if (isLogin) {
+      navigate('/')
+    }
+  }, [isLogin, navigate])
```
tadayosi, line 20: same as above.

packages/kubernetes-api-app/src/login/AuthLoginPage.tsx
tadayosi, line 1: Not related to the pull req itself, I found something weird in the codebase. Why do we have so many auth login pages in many places. I thought that the main reason we enhanced the plugin API at hawtio/react is for hawtio-online to use the API to extend the login page builtin in hawtio/react instead of having custom login pages like this.

If we don't use the login component extension here, there's no reason we keep having it :thinking:

packages/kubernetes-api-app/src/login/AuthLoginPage.tsx
phantomjinx: I did look at your comments. Although, not got to all of them. Rather than change the `useEffect` here, I was going to look at it in hawtio/next first and then consider changing them here after we had come to a conclusion. However, I can do it the opposite way around if preferred.

packages/kubernetes-api-app/src/login/AuthLoginPage.tsx
phantomjinx: There are 3 test apps.

1. oauth-app
2. kubernetes-app
3. management-app

These test apps do not use the UI of hawtio at all but isolate the functionality to levels below. So

1. oauth-app only tests authentication to the cluster;
2. kubernetes-app only tests up to the listing of resources in the cluster;
3. management-app only tests up to the listing of those pods with a jolokia connection and exercises those connections;

Therefore, the `AuthLoginPage` is provided in each case to replicate that used in hawtio but not actually use the one in hawtio, since the `Hawtio` component is intentionally not used in their user interfaces.

The `online-shell` package does use the hawtio UI and hence plugs into the core UI login page. This is the one that goes forward to being the actuall app.

I can create a test package that provides a single replicated `AuthLoginPage` rather than have the 3 of them but just hasn't been a priority up until now.

packages/kubernetes-api-app/src/login/AuthLoginPage.tsx
```diff
@@ -1,27 +1,40 @@
...
+
+  if (!userLoaded || (!hawtconfigLoaded && !pluginsLoaded)) {
```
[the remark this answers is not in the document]
phantomjinx: Updating the condition for the tests apps.

packages/kubernetes-api-app/src/login/AuthLoginPage.tsx
phantomjinx: To address the comment. Testing each of the test-apps with the `useEffect` removed leaves the page stuck on the `Loading...` screen. Despite logging-in the page never refreshes.

packages/kubernetes-api-app/src/login/AuthLoginPage.tsx
tadayosi: Ah so they are test code. Then I think it's fine at this moment. Thanks for explanation.

tadayosi (approved):
LGTM now.

Issue #210, opened 2023-11-22
Discover should show loading view instead of empty page for initial loading


tadayosi:
As you can see in the screencast, it first shows "No Hawtio containers" message. Is it difficult to keep showing [loading view](https://v4-archive.patternfly.org/v4/components/skeleton)?

[Screencast from 2023-11-22 12-57-59.webm](https://github.com/hawtio/hawtio-online/assets/156692/8a743b6e-ee79-458b-a75e-51060183818d)

PR #212, opened 2023-11-24
fix: Stops displaying no containers page while waiting for pods #210

phantomjinx:
* management-service.ts
  * The management service is responsible for emitting an update on all the pods being managed or not. However, no update is emitted if there are no pods. As a result, the Discover component has to assume there are no pods while waiting for them.
  * To allow for the Discover component to wait for the update (showing 'loading'), fire an update when there are no pods
  * Also for each pod add a flag to show that it has now been 'managed'

* context.ts
  * Differentiate between initialising / mounting of the context and the first call to organising the pods and the later mgmt service listener
  * Once mounted, only when the listener receives its first update will the context signal it is loaded (that signal will be either when there are pods or not hence the mandated firing from the mgmt service)

* Discover.tsx
  * Improves the loading component to show a header to the loading card

* DiscoverPodItem.tsx
  * In the event that a new pod is spun up, it can take several seconds for an update to be received concerning the number of routes. As a result, have the routes count label display a spinner and `querying ...` label before being updated to the correct total

### DIFF
--- a/packages/online-shell/src/discover/Discover.tsx
+++ b/packages/online-shell/src/discover/Discover.tsx
@@ -9,6 +9,10 @@ import {
   List,
   PageSection,
   PageSectionVariants,
+  Panel,
+  PanelHeader,
+  PanelMain,
+  PanelMainBody,
   Title,
 } from '@patternfly/react-core'
 import { CubesIcon } from '@patternfly/react-icons'
@@ -24,7 +28,18 @@ export const Discover: React.FunctionComponent = () => {
     useDisplayItems()
 
   if (isLoading) {
-    return <HawtioLoadingCard />
+    return (
+      <PageSection variant={PageSectionVariants.light}>
+        <Panel className='discover-loading'>
+          <PanelHeader>Waiting for Hawtio Containers ...</PanelHeader>
+          <PanelMain>
+            <PanelMainBody>
+              <HawtioLoadingCard />
+            </PanelMainBody>
+          </PanelMain>
+        </Panel>
+      </PageSection>
+    )
   }
 
   if (error) {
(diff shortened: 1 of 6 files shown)


phantomjinx:
https://github.com/hawtio/hawtio-online/assets/1634180/202bca38-38bf-432a-acd5-b68a8bfb77f6

tadayosi (approved):

packages/online-shell/src/discover/Discover.tsx
```diff
@@ -24,7 +28,18 @@ export const Discover: React.FunctionComponent = () => {
...
 
   if (isLoading) {
-    return <HawtioLoadingCard />
+    return (
+      <PageSection variant={PageSectionVariants.light}>
+        <Panel variant='raised' className='discover-loading'>
+          <PanelHeader>Waiting for Hawtio Containers ...</PanelHeader>
+          <PanelMain>
+            <PanelMainBody>
+              <HawtioLoadingCard />
```
tadayosi, line 37: Inside the panel, the loading card has a border and it doesn't look nice a bit.

tadayosi: Maybe we don't need the panel and header at all, and you can directly use `<HawtioLoadingCard>` for alignment with the other views.

phantomjinx: I can make the panel invisible but given the loading can take ~7 or 8 seconds, I felt that the text was essential to explain to the user what was loading exactly. We could place an optional text header in `HawtioLoadingCard` at a later point?

Anyway I'll make the panel invisible then do a screenshot and you can see what you think.

phantomjinx: Screenshot_20231127_105004

tadayosi: ok, lets go with it for now.

Issue #132, opened 2023-03-16
Migrate to quay.io


tadayosi:
Since Docker is sunsetting Free Team organisations:
https://web.docker.com/rs/790-SSB-375/images/privatereposfaq.pdf

tadayosi:
Docker free team orgs hasn't sunset but it's good time for us to switch to quay.io starting from v2. @phantomjinx

phantomjinx:
Completed with https://github.com/hawtio/hawtio-online/pull/229

PR #229, opened 2023-11-29
fix: Moves default image registry to quay.io #132

phantomjinx:
* deployment.yml
  * Updates image property to point to quay.io

* package.json
  * Updates companion commands to point to quay.io

* Makefile
  * Makes VERSION parameter overrideable from ENV
  * Updates IMAGE_NAME to quay.io
  * Adds VERSION override that updates SNAPSHOT suffix with date timestamp
  * Removes clutter of bundle-making as only applicable to operator

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,55 +19,32 @@
 #
 SHELL := /bin/bash
 
-VERSION ?= 2.0.0-SNAPSHOT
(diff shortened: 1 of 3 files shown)


tadayosi (approved):
thx

PR #232, opened 2023-12-03
Clear site data on logout

phantomjinx:
Since the header "Clear-Site-Data" is a server response header, the app must be directed to a landing page on logout that will allow this header to be communicated.

Provides a redirect page that when the user does a logout then are redirected first to the /logout url. This displays a "Going to provider login" page that countsdown then redirects after 5 seconds (or if the user clicks the link directly). This page provides the "Clear-Site-Data" header when called and clears the users storage for the domain.


phantomjinx:
https://github.com/hawtio/hawtio-online/assets/1634180/9fe022fd-236a-4959-aacc-ce674b769d15

docker/nginx-gateway.conf.template
```diff
@@ -71,6 +71,13 @@ server {
         internal;
     }
 
+    location = /logout {
+        add_header Clear-Site-Data "\"*\"";
+        add_header Cache-Control "no-store";
+        alias      /usr/share/nginx/html/online;
+        try_files  /redirectlogin.html /online/redirectlogin.html =504;
```
tadayosi, line 78: Can't we just immediately redirect to login here?

docker/nginx-gateway-k8s.conf.template
```diff
@@ -70,6 +70,13 @@ server {
         internal;
     }
 
+    location = /logout {
```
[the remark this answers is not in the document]
phantomjinx: So `/logout` was chosen simply because it was free. The `Clear-Site-Data` header is a server response so needed a concrete endpoint for the client to hit in order for the server to send the header. So as long as `/auth/logout` is not taken up by anything else then that should be fine.

docker/redirectlogin.html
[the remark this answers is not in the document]
phantomjinx: We can. It was useful as a way to ensure the /logout endpoint was reached. I'll try with a straight redirect then.

docker/nginx-gateway.conf.template
phantomjinx: Oauth does not always use the `/auth/login` endpoint due to OS providing its own. So need to ensure that the endpoint that the app wants to send is communicated to the server from the client.

packages/oauth/src/form/form-service.ts
```diff
@@ -18,20 +18,20 @@ interface Headers {
...
     this.userProfile = userProfile
     this.userProfile.setOAuthType(FORM_AUTH_PROTOCOL_MODULE)
     this.formConfig = formConfig
-    this.loggedIn = this.initLogin()
+    this.login = this.createLogin()
     this.fetchUnregister = null
   }
 
-  private initLogin(): boolean {
+  private async createLogin(): Promise<boolean> {
```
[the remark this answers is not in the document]
phantomjinx: Ran into problems with making the login form not display before the redirect had completed. This seemed to resolve an issue but may now be redundant so will re-assess.

phantomjinx:
So one thing to note ...

In order for nginx to be able to successfully redirect to the login provider page, a conversion has to be performed to decode the URI encapsulated in the `redirect_uri` query parameter. 'Pure' nginx is not capable of performing this so the only serviceable way of achieving this is to use a custom function in the `njinx.js` to call `decodeUriComponent()`. The result is then returned. This is not ideal but the only mechanism that does not require installing more modules in nginx.

packages/oauth/src/form/form-service.ts
```diff
@@ -176,7 +176,7 @@ export class FormService implements OAuthProtoService {
   }
 
   async isLoggedIn(): Promise<boolean> {
-    return await this.login
+    return this.login
```
tadayosi, line 179: Don't forget to remove `async` above.

tadayosi (approved):
One comment but otherwise it looks good to me.

packages/oauth/src/form/form-service.ts
phantomjinx: Darn! That sneaked in there!

phantomjinx: Removed `async` but Promise must remain to conform with interface since the OS OAuthProtoService is definitely asynchronous in the method.

Issue #200, opened 2023-11-17
Investigate disabling of connect plugin without resorting to hawtconfig.json


phantomjinx:
The Connect plugin should not be visible in hawtio-online since Discover takes on this functionality. Find a way to disable it more appropriately than resorting to the `hawtconfig.json` file that can be updated by the user.

phantomjinx:
Supercedes PR #197

tadayosi:
I don't see connection between this issue and #197. Do you mean this?
> * Corrects the route of Connect plugin in disabledRoutes
>   * Stops the Connect plugin from being displayed in Online-Shell app

PR #244, opened 2023-12-06
fix: Disable the connect plugin #200

phantomjinx:
* Stop the connect plugin from being displayed in the plugin bar by overriding its active function


tadayosi (approved):
There should be a better way, but let's move things forward with it for now. We can revisit it later.

phantomjinx:
> There should be a better way, but let's move things forward with it for now. We can revisit it later.

Yeah agreed. Again we would need to extend the API in hawtio-next I think.

PR #245, opened 2023-12-06
Adds other commands to deploy install Makefile

phantomjinx:
* deploy/Makefile
 * Adds generate-proxying, generate-serving & authclient as rules and includes them in the install

* Moves bash scripts to deploy/script and soft-links to them from root scripts directory to preserve working commands in yarn package.json

* generate-serving.sh
 * Updates script to include check for kube binary and the namespace for the secret


phantomjinx:
> Looks OK. Have you checked that the existing `yarn gen:proxying` and `yarn gen:serving` work fine with it?

Yes. I wanted to ensure current usability is maintained, especially as not sure when / if we might change over to promoting the use of the deploy/Makefile in the README rather than `yarn`.

PR #275, opened 2023-12-20
Stops display of the token if not JWT and if Openshift cluster fetches the username from the master API

phantomjinx:
Asks the Openshift cluster for the username associated with the given token.

If using form-login and connecting to Openshift, it is possible to obtain the username associated with the login token using the OS REST API.

Adds a small function to the `form-service.ts` that fetches the username upon successful login.

Screenshot_20231220_190358

### DIFF
--- a/packages/oauth/src/form/form-service.ts
+++ b/packages/oauth/src/form/form-service.ts
@@ -4,7 +4,7 @@ import $ from 'jquery'
 import { jwtDecode } from 'jwt-decode'
 import { relToAbsUrl } from 'src/utils/utils'
 import { OAuthProtoService, UserProfile, log } from '../globals'
-import { getCookie, logoutRedirect, redirect, secureDispose, secureRetrieve } from '../utils'
+import { FetchOptions, fetchPath, getCookie, joinPaths, logoutRedirect, redirect, secureDispose, secureRetrieve } from '../utils'
 import { FORM_AUTH_PROTOCOL_MODULE, FORM_TOKEN_STORAGE_KEY, FormConfig, ResolveUser } from './globals'
 
 type LoginOptions = {
@@ -15,6 +15,11 @@ interface Headers {
   Authorization: string
   'X-XSRF-TOKEN'?: string
 }
+
+/*
+ * OpenShift core API groups
+ */
+const OpenshiftAPIs = ['route.openshift.io', 'image.openshift.io', 'console.openshift.io']
 
 export class FormService implements OAuthProtoService {
   private userProfile: UserProfile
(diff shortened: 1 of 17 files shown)


phantomjinx:
Was experimenting with the cluster REST API and tried this out. Seems to work correctly but will probably need tidying up if we want to include it. Anyway, post what you think of it.

packages/oauth/src/form/form-service.ts
```diff
@@ -201,6 +233,16 @@ export class FormService implements OAuthProtoService {
         else log.error(err)
       }
 
+      //
+      // If token failed to be decoded then try and see if
+      // it is an openshift cluster and ask the API for the username
+      //
+      if (subject === this.userProfile.getToken()) {
+        // Will return either the username or null (if api request not valid)
+        // If null, subject will remain as the token
```
tadayosi, line 242: It's still wrong to use the token as the fallback for a user name. "subject" should mean the subject retrieved from a JWT token, not a general token itself. We should rewrite the entire logic and use it only for the subject from JWT, and just use a default name for the user such as `user`.

packages/oauth/src/form/form-service.ts
```diff
@@ -201,6 +233,16 @@ export class FormService implements OAuthProtoService {
         else log.error(err)
       }
 
+      //
+      // If token failed to be decoded then try and see if
+      // it is an openshift cluster and ask the API for the username
+      //
+      if (subject === this.userProfile.getToken()) {
```
tadayosi, line 240: Also we'd like to have a guard whether it's on OpenShift or not before sending an unnecessary request to the API server.

tadayosi (changes requested):
The username resolution logic looks good, but there are other comments that need to be addressed.

packages/oauth/src/form/form-service.ts
tadayosi: Why using token as user name is wrong is that it may expose the secret credential to the UI. It's like showing password to the UI. So taking screenshots for the UI may leak it. Users shouldn't need to take care about it.

tadayosi:
And while it's fine to support such case, it's important that it's just an edge case and should never impede the normal case with K8s and serviceaccount token by adding extra load.

packages/oauth/src/form/form-service.ts
```diff
@@ -180,6 +180,38 @@ export class FormService implements OAuthProtoService {
     return Promise.resolve(this.login)
   }
 
+  /**
+   * Tries to fetch the username connected to the token
+   * using the Openshift API.
+   * If the request fails, eg. cluster is not OS, then
+   * return null
```
tadayosi, line 187: So, in the light of this case being an edge case, it should be better to document here that it won't happen for most of times, but we are just supporting it mainly for our developing purposes.

phantomjinx:
> And while it's fine to support such case, it's important that it's just an edge case and should never impede the normal case with K8s and serviceaccount token by adding extra load.

Despite attempts to avoid an enquiry of the API Server to determine if it belongs to an OpenShift cluster, I have found no other way to get that result. The `form config` makes no mention of the type of cluster (we could include it but then it is dependent on the user correctly supplying it).

So the latest version:
- no longer displays the token as the subject;
- Evaluates the response of the base `/api` to determine if it contains the core OpenShift groups; 
- Only in the event that the cluster is OpenShift will it then enquire as to the username.

tadayosi (approved):

tadayosi:
@phantomjinx 
> Despite attempts to avoid an enquiry of the API Server to determine if it belongs to an OpenShift cluster, I have found no other way to get that result. The form config makes no mention of the type of cluster (we could include it but then it is dependent on the user correctly supplying it).

Just an idea to explore: What about using local storage and cache the result on whether it's OCP or K8s there?  The reasoning is that local storage is tied to a hostname and it's highly unlikely that the same hostname changes the type of cluster. If the result is cached, we only need to inquire once. If a similar inquiry has been done in other places in hawtio-online, we might also be able to use the cached result to remove extra burden on API server further.

tadayosi:
Yet another option might be that the nginx server holds the result of the inquiry [1] somewhere on the server in some way and provides an extra endpoint like `/cluster-type` which returns whether it's on OCP or K8s. That way we only need to inquire to the nginx server but not to the API server.

[1] https://github.com/hawtio/hawtio-online/blob/79baa7064957b3fa37d2ae14da1e25cdb490344f/docker/nginx.sh#L18-L31

PR #277, opened 2023-12-22
 fix: Add sanitization of the redirect uri to prevent XSS

mmelko:


packages/oauth/src/utils/urls.ts
```diff
@@ -50,8 +50,35 @@ export function logoutRedirect(redirectUri: URL): void {
...
+  return url.href
+}
+
 export function redirect(target: URL) {
   log.debug('Redirecting to URI:', target)
   // Redirect to the target URI
+  if (validateRedirectURI(target)) {
+    window.location.href = sanitizeUri(target)
+  } else {
+    log.error('Invalid redirectUrl parameter')
```
phantomjinx, line 81: I'd say if the redirect is invalid then perhaps throwing an Error would be beneficial since something really would have gone wrong. It could get lost in the error log. Indeed, we'd also want to stop the process from going any further, especially allowing the invalid url to be run unsanitized?

phantomjinx (changes requested):

oscerd (approved):

packages/oauth/src/utils/urls.ts
mmelko: I will throw the error. I believe we are essentially doing that. Uri will be unsatinized only if it doesn't have any parameters which is fine.  To proceed with redirect, the URL has to be valid at first, so in this case protocol host and port is the same as host.  (Have to be tested though)

mmelko: Updated

packages/oauth/src/form/login/form-auth-login-service.ts
```diff
@@ -58,7 +66,12 @@ class FormAuthLoginService {
     const currentUri = new URL(window.location.href)
     const searchParams: URLSearchParams = currentUri.searchParams
     if (searchParams.has('redirect_uri')) {
-      return searchParams.get('redirect_uri') as string
+      const uri = new URL(searchParams.get('redirect_uri') as string)
+      if (validateRedirectURI(uri)) {
+        return sanitizeUri(uri)
+      } else {
+        log.error('invalid redirect_uri')
```
tadayosi, line 73: Probably it's better to add the invalid `uri` to the log so a dev can know what went wrong with the URI.

packages/oauth/src/form/login/form-auth-login-service.ts
```diff
@@ -58,7 +66,12 @@ class FormAuthLoginService {
     const currentUri = new URL(window.location.href)
     const searchParams: URLSearchParams = currentUri.searchParams
     if (searchParams.has('redirect_uri')) {
-      return searchParams.get('redirect_uri') as string
+      const uri = new URL(searchParams.get('redirect_uri') as string)
+      if (validateRedirectURI(uri)) {
+        return sanitizeUri(uri)
+      } else {
+        log.error('invalid uri')
```
tadayosi, line 73: Looks like you misunderstood my feedback comment. What I was saying is something like:
```
log.error('invalid uri:', uri)
```
Does it make sense?

phantomjinx (approved):

PR #292, opened 2024-01-12
Fixes to dev server configuration

phantomjinx:
Talking with @mmelko, it became apparent that the original [commit](https://github.com/hawtio/hawtio-online/commit/3924aad861a42089c772b37a87d19d56feffaa36) that added the `/online` prefix to the `devServer` only seemed to half-work and failed entirely to find the `/login` page when the environment was configured for `form` mode. So this reverts that setup but keep `/online` as the prefix when the project is built for installation.

Encountered an error when testing on OpenShift that had the `authclient.sh` script being executed too soon.

### DIFF
--- a/deploy/Makefile
+++ b/deploy/Makefile
@@ -175,12 +175,6 @@ endif
 ifeq ($(CLUSTER_TYPE), openshift)
 	@$(MAKE) -s generate-proxying
 endif
-# Add the oauth client on openshift, if specified
-ifeq ($(CLUSTER_TYPE), openshift)
-ifeq ($(MODE), cluster)
-	@$(MAKE) -s cluster-auth-client
-endif
-endif
 # Add the tls serving on k8s, if specified
 ifeq ($(CLUSTER_TYPE), k8s)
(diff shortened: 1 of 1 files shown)


packages/online-shell/webpack.config.dev.js
```diff
@@ -227,7 +214,12 @@ module.exports = () => {
         // for accessing /login from the app. Having it here allows
         // the paths above to remain external to the app
         //
-        const history = historyApiFallback()
+        const history = historyApiFallback({
+          verbose: true,
+          index: '/',
+        })
+
+        console.log(history)
```
mmelko, line 222: do we want to have console.log ? or is it just a leftover

mmelko (commented):
one comment otherwise LGTM

mmelko:
Tested with CRC and works as expected.

packages/online-shell/webpack.config.dev.js
phantomjinx: leftover. Will remove.

tadayosi (approved):

PR #329, opened 2024-02-16
Encryption for local storage values

phantomjinx:
Provides fixes for the execution of any unit tests in each api package then implements the encryption / decryption of localStorage credentials.


packages/oauth/src/utils/secure-storage.ts
```diff
@@ -0,0 +1,60 @@
+import { decrypt, encrypt, generateKey, toBase64, toByteArray } from './crypto'
+
+const SESSION_KEY_SALT = 'connect.salt'
```
tadayosi, line 3: let's use a different name than the one from Connect plugin. if it's supposed to be used throughout Online, probably 'online.salt' is a better name.

tadayosi (approved):
one comment to be changed, but otherwise looks good to me.

Issue #363, opened 2024-03-06
Remove `HAWTIO_ONLINE_GATEWAY` env var and drop support for non-gateway Nginx deployment


tadayosi:
If I remember correctly, this gateway flag is provided to keep backward compatibility to non-RBAC hawtio-online deployment.
```yaml
            - name: HAWTIO_ONLINE_GATEWAY
              value: 'true'
```
But now, with v2 no one would use the old non-RBAC mode any more. To simplify the Nginx configurations, let's remove the env var and drop support for the non-gateway deployment mode.

phantomjinx:
https://github.com/hawtio/hawtio-online/pull/380.

Will also need change in operator to complete.

phantomjinx:
PR for operator: https://github.com/hawtio/hawtio-operator/pull/113

PR #380, opened 2024-03-13
fix: Simplies nginx configuration files (#363)

phantomjinx:
* Removes nginx.conf due to RBAC being compulsory

* Re-unites gateway conf files as only difference is the proxy cert/key properties and these are substituted in as vars

* Removes HAWTIO_ONLINE_GATEWAY property in deployment as no longer needed


oscerd (approved):

phantomjinx:
PR for operator: https://github.com/hawtio/hawtio-operator/pull/113

Issue #354, opened 2024-02-27
Hawtio Online should be robust in presense of unresponsive Jolokia endpoints


tadayosi:
When using Hawtio Online on K8s, _hawtio-enabled_ applications should disable client cert authentication for their jolokia endpoints. However, it's easily neglected and users might try to use Hawtio Online with client cert auth enabled applications on K8s. In such a scenario, the Online console becomes very irresponsive, due to unresponsive Jolokia endpoints at the beginning of probing pods. That's not a great experience for the users to start Hawtio Online. It should be robust enough to start quickly even in presense of some misconfigured pods.

Issue #379, opened 2024-03-12
User is spammed with login prompts on chrome


mmuzikar:
When Hawtio Online is making requests to the different jolokia containers, the requests get blocked with 403, which prompts users to put in username and password on chrome. This makes Hawtio Online basically unusable on chrome.

It might be related to some endpoints serving the index.html content, for example /keycloak/enabled or /auth/config both serve the HTML which fails on JSON parsing.

tadayosi:
@mmuzikar It could depend on the settings at the Jolokia containers side. Can you provide more info on your settings and if possible steps to reproduce the issue?

> for example /keycloak/enabled or /auth/config both serve the HTML which fails on JSON parsing.

I believe it's not a issue. It's just that they are a probe request and failures are expected, but the Fetch API always writes erros to browser console and I couldn't find a way to suppress the negligible errors.

mmuzikar:
I have the example app https://quay.io/repository/hawtio/hawtio-online-example-camel-springboot running on Openshift and running hawtio-online with yarn start:online and opening the website in Chrome. But the same behavior happens with hawtio-online normally running in the OCP cluster. I'm not aware of any changed settings.

phantomjinx:
This is an issue in testing in dev-mode using `yarn start:online`. It can be worked around by disabling the authentication requirements of jolokia on the test app using the following env vars:
- AB_JOLOKIA_AUTH_OPENSHIFT:               `false`
- AB_JOLOKIA_PASSWORD_RANDOM:        `false`
- AB_JOLOKIA_OPTS:                                      `useSslClientAuthentication=false,protocol=https`

We do point out in the [README](https://github.com/hawtio/hawtio-online/blob/main/README.md#disable-jolokia-authentication-for-deployments-dev-only) that this does need to be done.

phantomjinx:
Looking at this in terms of #354, I am adding better error handling & messaging to the Discover component. However, I looked to see if there was a way to disable this chrome login dialog and there is no good solution. Suggestions tend to be:

1. On the server change the response to something other than `401`
2. Modify the `WWW-Authenticate` header

It is the combination of these that chrome decides is the reason to display the dialog.

As a workaround, I am trying out an error count tracker for each pod which if it reaches a threshold will stop trying to probe the connection for changes. Not ideal but at least minimizes the number of dialogs to a more manageable number. However, this is only a workaround and not an ideal situation.

tadayosi:
[...]

PR #385, opened 2024-03-14
Handle jolokia connection failure and extending delay when connection errors

phantomjinx:
- fix: Handle jolokia probe failure (#354)
  - managed-pod.ts
    -  Extends the error property of Management to store more detail of errors, ie. messsage and http code
    - error property returned to undefined upon successful connection
    - Modifies the probe function to store the errors that may be generated
  - management-service.ts
    - Moves search into managed-pod
  - DiscoverPodConnect.tsx
    - Adds Error section to display error failure
    - Disables the connect button if a mgmt error has occurred
  - DiscoverPodItem.tsx
    - If error has occurred then display 'unknown routes' label

- fix: Extends the delay in polling for pods in error
  - If a pod cannot be connected to then it has a mgmt error. In that case
      don't keep polling it every 2 seconds but double the delay before
      trying to connect again.
  - managed-pod.ts
    - Adds error polling state for storing both the polling count and the
       threshold
  - management-service.ts
    - If the pod has a management error then increment the error poll count
       and its threshold to increasingly delay before trying again
    - Once it does poll and successfully connects then resets the error poll
       count

### DIFF
--- a/packages/management-api/src/managed-pod.ts
+++ b/packages/management-api/src/managed-pod.ts
@@ -8,6 +8,7 @@ import { log } from './globals'
 import jsonpath from 'jsonpath'
 import { k8Api, KubePod, k8Service, ObjectMeta, PodStatus, joinPaths, PodSpec } from '@hawtio/online-kubernetes-api'
 import { ParseResult, isJolokiaVersionResponseType, jolokiaResponseParse } from './jolokia-response-utils'
+import { eventService } from '@hawtio/react'
 
 const DEFAULT_JOLOKIA_OPTIONS: BaseRequestOptions = {
   method: 'post',
@@ -210,5 +211,14 @@ export class ManagedPod {
         failCb(this.getManagementError() as Error)
       },
     })
+  }
+
+  errorNotify() {
+    const name = this.getMetadata()?.name || '<unknown>'
+    const mgmtError = this.getManagementError()
+    if (mgmtError) {
+      const msg = `${name}: ${mgmtError.message}`
+      eventService.notify({ type: 'danger', message: msg })
+    }
   }
 }
--- a/packages/management-api/src/managed-pod.ts
+++ b/packages/management-api/src/managed-pod.ts
@@ -106,23 +106,23 @@ export class ManagedPod {
     return new Jolokia(options)
   }
 
-  get kind(): string | undefined {
+  getKind(): string | undefined {
     return this.pod.kind
   }
 
-  get metadata(): ObjectMeta | undefined {
+  getMetadata(): ObjectMeta | undefined {
     return this.pod.metadata
   }
 
-  get spec(): PodSpec | undefined {
+  getSpec(): PodSpec | undefined {
     return this.pod.spec
   }
 
-  get status(): PodStatus | undefined {
+  getStatus(): PodStatus | undefined {
     return this.pod.status
   }
 
-  get management(): Management {
+  getManagement(): Management {
     return this._management
   }
 
--- a/packages/management-api/src/managed-pod.ts
+++ b/packages/management-api/src/managed-pod.ts
@@ -1,8 +1,13 @@
-import Jolokia, { BaseRequestOptions } from 'jolokia.js'
+import Jolokia, {
+  BaseRequestOptions,
+  Response as JolokiaResponse,
+  VersionResponse as JolokiaVersionResponse,
+} from 'jolokia.js'
 import $ from 'jquery'
 import { log } from './globals'
 import jsonpath from 'jsonpath'
 import { k8Api, KubePod, k8Service, ObjectMeta, PodStatus, joinPaths, PodSpec } from '@hawtio/online-kubernetes-api'
+import { ParseResult, isJolokiaVersionResponseType, jolokiaResponseParse } from './jolokia-response-utils'
 
 const DEFAULT_JOLOKIA_OPTIONS: BaseRequestOptions = {
   method: 'post',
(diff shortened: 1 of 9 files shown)


packages/management-api/src/managed-pod.ts
```diff
@@ -15,13 +15,22 @@ export type Management = {
   status: {
     managed: boolean
     running: boolean
-    error: boolean
+    // Optional error object - reset to undefined on successful connection
+    error?: {
+      message: string // Error message
+      code: number // HTTP error code of the error
```
tadayosi, line 21: If we use comments, let's use the standard JSDoc to make them official doc info instead of staying in a temporal form.

packages/management-api/src/managed-pod.ts
```diff
@@ -108,6 +121,36 @@ export class ManagedPod {
     return this._management
   }
 
+  getManagementError(): Error | null {
+    if (!this._management.status.error) return null
+
+    return new Error(`(${this._management.status.error.code}): ${this._management.status.error.message}`)
```
tadayosi, line 127: Is this format intended? `(${this._management.status.error.code})` should be like: `(404):Not found` (the parentheses)

packages/management-api/src/managed-pod.ts
```diff
@@ -108,6 +121,36 @@ export class ManagedPod {
...
+  }
+
+  private setManagementError(errorCode: number, errorMsg: string) {
+    this._management.status.error = {
+      code: errorCode,
+      message: errorMsg,
+    }
+  }
```
tadayosi, line 135: Just FYI, you can write (more JS like style :wink: ):
```
  private setManagementError(code: number, message: string) {
    this._management.status.error = { code, message }
  }
```

packages/management-api/src/managed-pod.ts
```diff
@@ -108,6 +121,36 @@ export class ManagedPod {
...
+
+  get errorPolling(): ErrorPolling {
+    return this._errorPolling
+  }
```
tadayosi, line 139: When we only intend to support getter, using JS getter/setter notation is a bit too much without adding any values. Later we might want to make it promise but then it can become a bit painful to rewrite it if it uses the getter/setter notation.

My experiences with Hawtio development now tells me that we should avoid the JS notation altogether.

packages/management-api/src/managed-pod.ts
```diff
@@ -108,6 +121,36 @@ export class ManagedPod {
...
+
+  resetErrorPolling() {
+    this._errorPolling.count = 0
+    this._errorPolling.threshold = 1
+  }
+
+  incrementErrorPollCount() {
+    this._errorPolling.count = ++this._errorPolling.count
```
tadayosi, line 147: Can't we just make it like this?  It looks a bit redundant.
```
this._errorPolling.count++
```

packages/management-api/src/managed-pod.ts
```diff
@@ -117,36 +160,60 @@ export class ManagedPod {
...
-            }
-
-            if ('value' in resp && 'agent' in resp.value) {
-              log.debug('Found jolokia agent at:', this.jolokiaPath, 'version:', resp.value.agent)
-              resolve(this.jolokiaPath)
-              return
-            } else {
-              throw new Error('Detected jolokia but cannot determine agent or version')
-            }
+            jsonResponse = JSON.parse(data)
```
tadayosi, line 170: Basically, if we know it's a response from Jolokia we should use the type provided by `jolokia.js`. We are modern code.
https://github.com/jolokia/jolokia/blob/af7075546281f65da13cd42b075bc03670f2535d/client/javascript/src/main/javascript/index.d.ts#L504-L516

tadayosi (approved):
LGTM in general. There are a few feedback comments.

packages/management-api/src/managed-pod.ts
```diff
@@ -117,36 +160,60 @@ export class ManagedPod {
...
+        .fail((xhr: JQueryXHR, _: string, error: string) => {
+          const msg = `Jolokia Connect Error (${xhr.status}): ${error ?? xhr.statusText}`
+          this.setManagementError(xhr.status, msg)
+          reject(this.getManagementError())
         })
     })
   }
+
+  search(successCb: () => void, failCb: (error: Error) => void) {
+    this.jolokia.search('org.apache.camel:context=*,type=routes,*', {
```
tadayosi, line 206: Just in case, what is this method for?  Note a pod may not necessarily contain a Camel context. The only requirement is that a pod should be _hawtio-enabled_ which means it has a Jolokia port, not necessarily to be a Camel app. For example, we should also support Artemis brokers which don't have Camel contexts at all.

packages/management-api/src/managed-pod.ts
phantomjinx: Yes. I though it made use of the http code and seems to be consistent with the way HTTP Errors are displayed when converted to strings. Happy to change if preferred.

packages/management-api/src/managed-pod.ts
phantomjinx: So would you prefer the more java-like style, eg. `getErrorPolling()`, or just making the vars 'public' (well not private - JS scoping tends to be sugar coating at times)?

Certainly, there have been occasions when I prefer creating setters (& rarely getters) when some extra operation is required, eg. the incrementPollCounter(), so I presume that kind of style is ok?

packages/management-api/src/managed-pod.ts
phantomjinx: :-)
Yes we can. Although, I ran into trouble with that syntax as I started doing `const count = this._errorPolling.count++` and count remained as 0! So, I think I was just being over-zealous in ensuring the value was indeed incremented.

packages/management-api/src/managed-pod.ts
tadayosi: OK, that's fine if it's intentional.

packages/management-api/src/managed-pod.ts
tadayosi: Good point. For now, we can choose the explicit getter method: `getErrorPolling()`. I personally see little value in using the Ruby-like implicit getter/setter notation in JS.

And if you need a setter, the style you mention is totally justifiable.

packages/management-api/src/managed-pod.ts
tadayosi: Ah I see. Thanks for the explanation. It's then totally a different topic, but maybe later we can improve this part (of course, with another pull req than this one) by making it display something like `non camel` or removing the `route` label when it's really not a Camel app.

phantomjinx:
Take note of this [comment ](https://github.com/hawtio/hawtio-online/issues/379#issuecomment-2002991065)

phantomjinx:
Revision of error handling in Discover component
- Alert is displayed on display of Discover for each pod error
- Clicking the Notification Badge will re-display the pod error

Screenshot_20240319_212543

PR #388, opened 2024-03-15
Unify loading components

### LINKED ISSUE #387: Use the same loading page with hawtio/react
Instead of:
loading1

use this builtin loading page for consistency:
loading2

The component is here:
https://github.com/hawtio/hawtio-next/blob/main/packages/hawtio/src/ui/page/HawtioLoadingPage.tsx

It doesn't seem to be exported so we need to export it first.

mmuzikar:
Closes #387


tadayosi (approved):
thx!

Issue #397, opened 2024-03-20
Hawtio-Online Discover component stuck on 'Waiting for Hawtio Containers ....'


phantomjinx:
Use-Case:
1. Open Hawtio-Online and wait for Discover to display all pods;
2. Click on User -> Preferences and display the Preferences page;
3. Click on Discover to restore the Discover page.

On first loading, the 'Waiting for Hawtio Containers ...' loading page is refreshed to display all the pods. However, after step `3`, the loading page remains and is never refreshed.

PR #398, opened 2024-03-20
fix: Sorts out the updating of the Discover component (#397)

phantomjinx:
* managed-pod.ts
  * Adds the fingerprint functions from management-service so that the fingerprint can be cached on the pod.
  * Important!! fingerprints the management properties and the original kube pod so that changes to the actual pod can be picked up

* management-service.ts
  * Adds the fireUpdate property to the updateQueue so that the queue knows to fire an update if ANY of the pods need to fire an update.
  * Therefore if the QUEUE not just the last pod needs to fire an update then it should be fired.

* context.ts
  * Merges the 2 useEffects into 1 that both organises pods immediately & upon a management update

* Use Cases:
  - Discover is correctly populated on startup
  - Discover is correctly populated when re-clicked on
  - Discover is correctly updated if a change on a pod occurs in cluster
  - Filter toolbar is retained even if a filter is added that filters out all pods
  - Filter can be removed and all pods correctly updated
  - Filters can be cleared and all pods correctly updated
  - Error properly displayed

### DIFF
--- a/packages/management-api/src/management-service.ts
+++ b/packages/management-api/src/management-service.ts
@@ -1,5 +1,5 @@
 import { EventEmitter } from 'eventemitter3'
-import { ManagedPod, Management } from './managed-pod'
+import { ManagedPod } from './managed-pod'
 import { Connection, Connections, connectService } from '@hawtio/react'
 import { k8Service, k8Api, KubePod, K8Actions, Container, ContainerPort, debounce } from '@hawtio/online-kubernetes-api'
 import { MgmtActions, log } from './globals'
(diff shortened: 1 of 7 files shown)


packages/management-api/src/management-service.ts
```diff
@@ -122,7 +121,7 @@ export class ManagementService extends EventEmitter {
          * against a non-running pod or a pod that cannot connect via jolokia
          * Emit an update but only if the status has in fact changed
          */
-        this.emitUpdate({ uid, fireUpdate: fingerprint !== this.fingerprint(mPod.getManagement()) })
+        this.emitUpdate({ uid, fireUpdate: mPod.getFingerprint() !== mPod.calcFingerprint() })
```
tadayosi, line 124: if these are the only places where the fingerprint is used, we can put the fireUpdate logic into the managed pod class instead of repeating it here. we might be also able to make the fingerprint property and the getter/calc methods private, which further contributes to making the managed pod api simpler.

tadayosi (approved):
a few minor comments but otherwise looks good to me.

PR #408, opened 2024-03-27
fix(ux): redirect to origin rather than the same url

### LINKED ISSUE #407: User is left on a blank page after logging off
When a user logs off from being connected to an app they get redirected to login screen again and if they log in they get navigated back to hawtio online, but there's no content presented. The issue is probably that the same redirect_uri is used, but I guess the users should be redirected back to just /online not /online/<path>

[Peek 2024-03-27 15-32.webm](https://github.com/hawtio/hawtio-online/assets/46345469/c76eda88-5506-441d-b790-03d180e9f67e)

mmuzikar:
Closes #407

### DIFF
--- a/packages/oauth/src/openshift/osoauth-service.ts
+++ b/packages/oauth/src/openshift/osoauth-service.ts
@@ -1,4 +1,4 @@
-import { PUBLIC_USER, ResolveUser, userService } from '@hawtio/react'
+import { hawtio, PUBLIC_USER, ResolveUser, userService } from '@hawtio/react'
 import * as fetchIntercept from 'fetch-intercept'
 import $ from 'jquery'
 import { OAuthProtoService } from '../api'
@@ -283,14 +283,15 @@ export class OSOAuthService implements OAuthProtoService {
   private doLogout(config: OpenShiftOAuthConfig): void {
     this.fetchUnregister?.()
 
-    const currentURI = new URL(window.location.href)
+    const currentURL = new URL(`${window.location.origin}/${hawtio.getBasePath() || ''}`)
+
     // The following request returns 403 when delegated authentication with an
     // OAuthClient is used, as possible scopes do not grant permissions to access the OAuth API:
     // See https://github.com/openshift/origin/issues/7011
     //
     // So little point in trying to delete the token. Lets do in client-side only
     //
-    forceRelogin(currentURI, config)
+    forceRelogin(currentURL, config)
   }
 
   isLoggedIn(): Promise<boolean> {


packages/oauth/src/openshift/osoauth-service.ts
```diff
@@ -283,14 +283,13 @@ export class OSOAuthService implements OAuthProtoService {
...
 
-    const currentURI = new URL(window.location.href)
     // The following request returns 403 when delegated authentication with an
     // OAuthClient is used, as possible scopes do not grant permissions to access the OAuth API:
     // See https://github.com/openshift/origin/issues/7011
     //
     // So little point in trying to delete the token. Lets do in client-side only
     //
-    forceRelogin(currentURI, config)
+    forceRelogin(new URL(window.origin), config)
```
tadayosi, line 292: Does `window.online` really return the intended URL?  It doens't seem to include `/<path>` part. Isn't it better to use `hawtio.getBasePath()`, which is the standard Hawtio API to get the console's base path?

tadayosi (commented):
One comment. Not sure if `window.origin` is the right choice.

packages/oauth/src/openshift/osoauth-service.ts
mmuzikar: `hawtio.getBasePath()` can return undefined and I thought it will return only `/online`

tadayosi: You are right, so you should use the combination like:
https://github.com/hawtio/hawtio-next/blob/7acdf61d5f4839f26cd33f35b81560fcb74aa3ae/packages/hawtio/src/plugins/shared/jolokia-service.ts#L421-L422

Are you sure that `window.origin` returns `/online`?  I couldn't find resources that assure so.
https://developer.mozilla.org/en-US/docs/Web/API/origin
https://stackoverflow.com/questions/55451493/what-is-window-origin
To me it looks like it returns something like `https://<cluster-hostname>:<port>`, just like `Origin:` http header.

tadayosi: I can only find: https://html.spec.whatwg.org/multipage/browsers.html#origin

tadayosi: `window.location.origin` seems to be safer: https://developer.mozilla.org/en-US/docs/Web/API/Location/origin in combination with `hawtio.getBasePath()`

packages/oauth/src/openshift/osoauth-service.ts
```diff
@@ -283,14 +283,13 @@ export class OSOAuthService implements OAuthProtoService {
...
 
-    const currentURI = new URL(window.location.href)
     // The following request returns 403 when delegated authentication with an
     // OAuthClient is used, as possible scopes do not grant permissions to access the OAuth API:
     // See https://github.com/openshift/origin/issues/7011
     //
     // So little point in trying to delete the token. Lets do in client-side only
     //
-    forceRelogin(currentURI, config)
+    forceRelogin(new URL(`${window.location.origin}/${hawtio.getBasePath()}`), config)
```
tadayosi, line 292: Shouldn't it lead to `https://host:port/undefined` if it doesn't handle undefined correctly?  I suggest to declare the URL in a separate line just like the original code.

mmuzikar:
No longer needed as the issue is no longer present

PR #439, opened 2024-04-18
fix: HAWNG-475 - Limits access to the cluster API using a whitelist

phantomjinx:
* nginx-gateway.conf.template
  * Changes the master location to first send it through a javascript guard function that determines if the URI is acceptable. Thereby disallowing requests such as secrets and configmaps.

* nginx.js
  * proxyMasterGuard function that checks the given uri against a set of accepted patterns.
    * If pattern not found then return a 502 and error message.
    * If found then internally redirect to the masterinternal location

* test.js
  * Adds tests to check syntax and exercise proxyMasterGuard

### DIFF
--- a/docker/nginx-gateway.conf.template
+++ b/docker/nginx-gateway.conf.template
@@ -105,9 +105,26 @@ server {
         gzip_static on;
     }
 
-    # Kubernetes master API reverse proxying
+    #
+    # External location which checks whether the
+    # requested uri is accessible before redirecting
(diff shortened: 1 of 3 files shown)


phantomjinx:
Trying to access the cluster secrets through the proxy API from external app returns a 502 error and a 'not allowed' message:

Screenshot_20240417_210124

tadayosi (approved):

PR #465, opened 2024-05-13
HAWNG-666: jolokia search function not included in imports

phantomjinx:
* When using Jolokia class, need to import the simple API functions from jolokia.js/simple as well as the core jolokia.js


mmelko (approved):
lgtm

PR #491, opened 2024-07-02
Migration of njs to node container (gateway)

phantomjinx:
* Update of git actions to reflect second image

* Upgrade of yarn to 4.3.1
  * not necessary and could be broken out if too confusing

* Upgrade of node to version 20 
  * whether we should go whole way to version 22 ?

* Hawtio-Online Docker file renamed to `Dockerfile-nginx`
* Addition of `Dockerfile-gateway` for buidling of gateway image
* Inclusion of hawtio-403.html page
  ** Locations in nginx are now localhost-only rather than internal so can try them but will return a 403 error
 
* `Makefile` updated to include build of gateway image
* `deploy/Makefile` update to include install of gateway image

* `docker/gateway` project for the gateway image
  ** Express server with proxying capabilities to `master` (required due to websockets)
  ** Unit tests migrated from njs testing
  ** Logging with pino
  ** Development build with `tsup`
  ** Production build with webpack (switched from `tsup` due to persistent bundling errors)
  ** `js-yaml` replaced with `yaml` and imported as dependency
  ** `jwt-decode` imported as dependency

* HAWTIO_ONLINE_GATEWAY_APP_PORT exposed as env var allowing users to change running port if required (maybe not required in production but only in development)

* root `package.json` commands updated to consider `docker/gateway`


phantomjinx:
A review of 60 files may not be the easiest solution so if anyone would like to suggest an alternative then happy to accommodate.

Am sure there are issues so although a PR, lets consider this a working PR that I'll continue to update until all concerns met.

.github/workflows/image.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - njs-migrate
```
tadayosi, line 7: not need this anymore

.github/workflows/image.yml
```diff
@@ -20,14 +21,17 @@ jobs:
...
       - name: Push image to Quay.io
+        if: github.repository == 'hawtio/hawtio-online'
         env:
           USERNAME: ${{ secrets.QUAY_USERNAME }}
           PASSWORD: ${{ secrets.QUAY_PASSWORD }}
         run: |
-          docker login -u $USERNAME -p $PASSWORD quay.io
-          docker push quay.io/hawtio/online:latest
+          docker login -u ${USERNAME} -p ${PASSWORD} quay.io
+          CUSTOM_VERSION=2.1.0.202406281530 make image-push
```
tadayosi, line 37: why this custom version?

.github/workflows/lint.yml
```diff
@@ -9,6 +9,7 @@ on:
   push:
     branches:
       - main
+      - njs-migrate
```
tadayosi, line 12: no need for this anymore

.github/workflows/test.yml
```diff
@@ -9,6 +9,7 @@ on:
   push:
     branches:
       - main
+      - njs-migrate
```
tadayosi, line 12: same

Dockerfile-gateway
```diff
@@ -0,0 +1,46 @@
...
+COPY docker/gateway/tsconfig.json ./
+COPY docker/gateway/src ./src/
+COPY docker/gateway/public ./public/
+
+RUN yarn install
+RUN yarn build
+
+#====================================================
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.10
```
tadayosi, line 20: shouldn't we use ubi9?

Dockerfile-nginx
```diff
@@ -12,15 +12,13 @@ RUN yarn install
 RUN yarn build
 
 # Build stage to extract envsubst
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.8 as envsubst
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.10 as envsubst
```
tadayosi, line 15: same

deploy/base/deployment.yml
```diff
@@ -45,16 +45,49 @@ spec:
...
+              value: https://localhost:8443
+            - name: HAWTIO_ONLINE_GATEWAY_SSL_KEY
+              value: /etc/tls/private/serving/tls.key
+            - name: HAWTIO_ONLINE_GATEWAY_SSL_CERTIFICATE
+              value: /etc/tls/private/serving/tls.crt
+            - name: HAWTIO_ONLINE_GATEWAY_SSL_CERTIFICATE_CA
+              value: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+          ports:
+            - name: express
+              containerPort: 3443
```
tadayosi, line 75: is the app port also https? otherwise, 3443 is a bit misleading. maybe you can just use 3000?

docker/gateway/package.json
```diff
@@ -0,0 +1,78 @@
+{
+  "name": "@hawtio/gateway",
+  "version": "1.0.0",
```
tadayosi, line 3: maybe we can start with 2.1.0 instead to avoid confusions?

tadayosi (approved):
overall, it looks great.

PR #513, opened 2024-09-03
Fix console links and remove inconsistent link styling

### LINKED ISSUE #503: Pod chips don't redirect to href on link
The chips under the pod used to link users to the openshift console filtering for that specific chip, now it doesn't work. The href parameter is correct but the underlying element is button, so my guess is button doesn't respect the href attribute.

image

mmuzikar:
Patternfly requires for Buttons to have the `component='a'` attribute to respect the `href` attribute https://www.patternfly.org/components/button#links-as-buttons
Removes the underscore from the preferences dropdown of the first link
Closes #503


tadayosi (approved):

phantomjinx:
@mmuzikar
Can you rebase this, please?
I have upgraded webpack so should no longer error on the audit task in the test building.

Issue #496, opened 2024-07-16
Paginate the discover panel to only display a required number of pods at a time


phantomjinx:
Rather than displaying all pods found in the cluster that contain a jolokia port (which could be 100s!), only return a limited set, where the limit is selected by the user. Support the ability to then fetch the next set of pods while keeping track of the previous set (not the actual pods) so that the user can re-fetch the previous pods if required. At any time only the finite set of pods that the user has specified should be resident in client memory.

PR #525, opened 2024-09-10
Solution for paginating the Discover page (#496)

phantomjinx:
https://github.com/user-attachments/assets/5a85e823-08c2-4505-a660-44e08a887018


phantomjinx:
Will do some videos to show exactly how it works ... to follow.

tadayosi (commented):
It's a bit tough to review such big changes effectively. Could you outline what's the core idea of the solution and where is the most important part of the changes?

tadayosi:
Specifically, what's your solution to #496, how do you think the issue is solved with it?

phantomjinx:
> It's a bit tough to review such big changes effectively. Could you outline what's the core idea of the solution and where is the most important part of the changes?

Yeah, no problem. As I say, I'll do a video and talk through the changes

phantomjinx:
> Specifically, what's your solution to #496, how do you think the issue is solved with it?

### Changes
- Paging of pods (previous/next)
- Skipping of pods (first/last/page)
- Filter Namespace (filters out namespaces down to k8s level)
- Filter Pods (filters out pods down to k8s level)
- Sort (down to k8s level)
- Preference for controlling number of pods displayed in each page
- Preference for controlling the jolokia probe interval for checking on pod connectivity

### Specifics
1. When a namespace is scanned, all the pods that contain a jolokia connection are indexed by name;
2. Using a given namespace limit (default 3), the first 'page' of pods are fetched and watcher connections established. Therefore, for each namespace containing jolokia pods, only the first '3' pods are ever being watched at any 1 time;
3. When the user click 'next' or 'prev', the watchers for the current set of pods are destroyed and a new set is created for the 'next' set of 3. Therefore, regardless of the project size the UI only displays a set limited number;
4. A preference is provided, should the user wish to increase the display from 3;
5. Both filter and sort have been refactored to perform those functions down at the k8s level rather than just the UI level. This makes sense since the filter then applies to all pods in the namespace rather than just those displayed in the current page of the namespace in the UI.

packages/kubernetes-api/src/client/namespace-client.ts
```diff
@@ -0,0 +1,400 @@
...
+export type NamespaceClientCallback = (jolokiaPods: KubePod[], fullPodCount: number, error?: Error) => void
+
+export interface Client<T extends KubeObject> {
+  watched: Watched<T>
+  watch: ProcessDataCallback<T>
+}
+
+interface PodWatcher {
+  client: Client<KubePod>
+  jolokiaPod: KubePod | undefined
```
tadayosi, line 19: this can be simply:
```
  jolokiaPod?: KubePod
```

packages/kubernetes-api/src/client/namespace-client.ts
tadayosi, line 1: maybe we can consider adding unit tests for newly introduced classes like this, if it's not so complex to do so.

packages/kubernetes-api/src/client/support.ts
tadayosi, line 1: these are a good place to add some more tests.

packages/kubernetes-api/src/client/ws-handler.ts
```diff
@@ -29,16 +29,16 @@ export class WSHandlerImpl<T extends KubeObject> implements WSHandler<T> {
     return this._list || new ObjectListImpl<T>()
   }
 
-  get collection() {
-    return this._collection
+  get watched() {
```
tadayosi, line 32: better to define the returned type

packages/kubernetes-api/src/client/ws-handler.ts
```diff
@@ -189,41 +189,47 @@ export class WSHandlerImpl<T extends KubeObject> implements WSHandler<T> {
...
-          const wsURL = this.collection.wsURL
+          const wsURL = this.watched.wsURL
           if (wsURL) {
-            log.debug('Connecting websocket for kind:', this.collection.kind)
+            log.debug('Connecting websocket for kind:', this.watched.kind)
             this.socket = this.createWebSocket(wsURL)
             this.setHandlers(this, this.socket)
           } else {
-            log.info('No wsURL for kind: ' + this.collection.kind)
+            log.info('No wsURL for kind: ' + this.watched.kind)
```
tadayosi, line 219: you can use `,` instead of `+`

packages/kubernetes-api/src/client/ws-handler.ts
```diff
@@ -189,41 +189,47 @@ export class WSHandlerImpl<T extends KubeObject> implements WSHandler<T> {
...
           }
         }
 
-        log.debug('Fetching initial collection of object from web socket: ' + this.collection.restURL)
-        fetchPath(this.collection.restURL, {
+        log.debug('Fetching initial collection of object from web socket: ' + this.watched.restURL)
```
tadayosi, line 223: prefer `,` over `+`

packages/kubernetes-api/src/kubernetes-service.ts
```diff
@@ -289,4 +294,118 @@ export class KubernetesService extends EventEmitter {
...
+  }
+
+  page(pageIdx: number, namespace?: string) {
+    if (!namespace) return
+
+    const nsClient = this.namespace_clients[namespace]
+    if (!nsClient) return
+
+    if (!nsClient.isConnected) {
+      log.warn(`k8 Service cannot page on disconnected namespace ${namespace}`)
```
tadayosi, line 405: `k8` -> `k8s`

packages/kubernetes-api/src/paging-metadata-impl.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  /**
+   * Moves current back 1 to get the previous paging metadata
+   */
+  previous() {
+    if (this._current === 0) return
+
+    if (!this.hasPrevious()) return
+
+    --this._current
```
tadayosi, line 138: what's the difference between `--this._current` vs `this._current--`?  if there's no difference, then I'd prefer the latter.

packages/kubernetes-api/src/paging-metadata-impl.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  /**
+   * Moves current forward 1 to get the next paging metadata
+   */
+  next() {
+    if (this._current === this.numPages() - 1) return
+
+    if (!this.hasNext()) return
+
+    ++this._current
```
tadayosi, line 156: same here

packages/management-api/src/index.ts
```diff
@@ -15,5 +15,6 @@ export async function isMgmtApiRegistered(): Promise<boolean> {
   return await registerManagementApi()
 }
 
-export { MgmtActions, log } from './globals'
 export { ManagedPod } from './managed-pod'
+export { ManagedProject } from './managed-project'
+export * from './globals'
```
tadayosi, line 20: maybe we should not export everything from globals, as elements such as pluginName and log are not relevant in other contexts. Only export the required types.

packages/online-shell/src/discover/discover-project.ts
tadayosi, line 1: another class maybe you can add tests.

tadayosi (approved):
GJ

packages/kubernetes-api/src/paging-metadata-impl.ts
phantomjinx:
```
If used postfix, with operator after operand (for example, x--), the decrement operator decrements and returns the value before decrementing.

If used prefix, with operator before operand (for example, --x), the decrement operator decrements and returns the value after decrementing.
```

In this context, since there is no variable then its make no difference. Prefixing has just been my default. Will update.

phantomjinx: As it happens, that is a class that is no longer used (was conceived with a different solution to the paging problem which I ultimately ditched).

packages/kubernetes-api/src/paging-metadata-impl.ts
phantomjinx: See comment above. No longer applicable.

PR #544, opened 2024-10-11
fix: Upgrade of hawtio/react to 1.5.0

phantomjinx:
* With this upgrade, the transport method of jolokia requests has been changed from jquery to fetch. This meant a number of changes in the development server in order to continue working.

* management-service.ts
 * Rather than just blindly connect to a target app, try testing the connection first and if it fails do not change location / open a new browser tab but notify the user with the event service

* oauth/[....]service.ts
 * Mandate a Content-Type of application/json when using fetch. Due to CORS this is not possible when actually calling fetch(uri, { headers })
 * Ensure when regsitering the fetch interceptor that the headers that have been created are not overwritten by any headers already contained in the requestConfig.

* webpack.config.dev.js
 * Re-implements management endpoint to conduct a sub-request using fetch to the master endpoint rather than redirecting. The fetch function appears to not redirect correctly when a Content-Length header is specified
 * Implementing a sub-request is the same way that the gateway component solves this issue
 * Ensures the token is passed to the sub-request
 * Logs whether the request body has been incorrectly parsed empty by the Express server in the webpack-dev-server.

### DIFF
--- a/packages/kubernetes-api-app/package.json
+++ b/packages/kubernetes-api-app/package.json
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "@hawtio/online-kubernetes-api": "workspace:*",
-    "@hawtio/react": "^1.3.0",
+    "@hawtio/react": "^1.5.0",
     "@patternfly/react-core": "^5.3.3",
     "@patternfly/react-styles": "^5.3.1",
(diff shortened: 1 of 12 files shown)


phantomjinx:
Will supercede #537

Issue #440, opened 2024-04-18
Consider methods to consolidate headers in nginx configuration


phantomjinx:
To conform with security requirements, the number of add_header config lines is proliferating in the nginx config template. One of the reasons for the number is that adding a single header in a location will not merge with the parent's set of headers but overwrite them!

Therefore, all parent headers must be specified again in a child location if the latter has an additional add_header entry. 

There are workarounds to this that could be considered:
- [most_set_headers](https://www.getpagespeed.com/server-setup/nginx/the-pitfalls-of-add_header-in-nginx-solving-inheritance-issues-with-more_set_headers) directive - provides an alternative merge-style directive to be used instead of add_header;
- [include](https://www.peterbe.com/plog/be-very-careful-with-your-add_header-in-nginx) directive - putting all the add_header entries into a single file then include that file with a single-line include directive.

PR #550, opened 2024-10-15
fix: Reduces repetition of headers in nginx config file (#440)

phantomjinx:
* Rather than repeating add_header directives, use 1-line include directives instead to bring import an additional file

* Makes the include files themselves hierarchical with further includes back to a single base include file


tadayosi (approved):
LGTM

PR #552, opened 2024-10-17
fix(#502): extend CSP to work with monaco-editor

### LINKED ISSUE #502: Monaco fails loading on Hawtio Online
When accessing the route source code monaco fails to initialize with the following errors:
```
Content-Security-Policy: The page’s settings blocked a script (script-src-elem) at https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/vs/loader.js from being executed because it violates the following directive: “default-src 'self'” index.js:86:23
Monaco initialization: error: 
error { target: script, isTrusted: true, srcElement: script, eventPhase: 0, bubbles: false, cancelable: false, returnValue: true, defaultPrevented: false, composed: false, timeStamp: 8824, … }
​bubbles: false
​cancelBubble: false
​cancelable: false
​composed: false
​currentTarget: null
​defaultPrevented: false
​eventPhase: 0
​explicitOriginalTarget: <script src="https://cdn.jsdelivr.net…@0.43.0/min/vs/loader.js">
​isTrusted: true
​originalTarget: <script src="https://cdn.jsdelivr.net…@0.43.0/min/vs/loader.js">
​returnValue: true
​srcElement: <script src="https://cdn.jsdelivr.net…@0.43.0/min/vs/loader.js">
​target: <script src="https://cdn.jsdelivr.net…@0.43.0/min/vs/loader.js">
​timeStamp: 8824
​type: "error"
​<get isTrusted()>: function isTrusted()
​<prototype>: EventPrototype { composedPath: composedPath(), stopPropagation: stopPropagation(), stopImmediatePropagation: stopImmediatePropagation(), … }
index.mjs:1:4528
```

mmelko:
closes #502

### DIFF
--- a/docker/includes/security-headers-online.conf
+++ b/docker/includes/security-headers-online.conf
@@ -3,4 +3,4 @@ include /etc/nginx/includes/security-headers-base.conf;
 # Govern what content can be loaded by the server and from where
 # Click jacking prevention to be used in addition to X-Frame-Options
 # Requires allowing inline-styles and inline data objects (svg imgs)
-add_header Content-Security-Policy "default-src 'self'; style-src 'self' 'unsafe-inline'; font-src 'self' data:; img-src 'self' data:; frame-ancestors 'self'; form-action 'self'; ";
+add_header Content-Security-Policy "default-src 'self'; style-src 'self' 'unsafe-inline'; font-src 'self' data:  https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/; img-src 'self' data:; frame-ancestors 'self'; form-action 'self'; script-src-elem 'self' https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/; style-src-elem 'self' 'unsafe-inline' https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/;";


docker/includes/security-headers-online.conf
```diff
@@ -3,4 +3,4 @@ include /etc/nginx/includes/security-headers-base.conf;
 # Govern what content can be loaded by the server and from where
 # Click jacking prevention to be used in addition to X-Frame-Options
 # Requires allowing inline-styles and inline data objects (svg imgs)
-add_header Content-Security-Policy "default-src 'self'; style-src 'self' 'unsafe-inline'; font-src 'self' data:; img-src 'self' data:; frame-ancestors 'self'; form-action 'self'; ";
+add_header Content-Security-Policy "default-src 'self'; style-src 'self' 'unsafe-inline'; font-src 'self' data:  https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/; img-src 'self' data:; frame-ancestors 'self'; form-action 'self'; script-src-elem 'self' https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/; style-src-elem 'self' 'unsafe-inline' https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/;";
```
tadayosi, line 6: Bad news is that now monaco-editor for hawtio-react 1.5.0 is 0.52.0. Doesn't it mean we should use 0.52.0?
```console
$ yarn why monaco-editor
├─ @hawtio/react@npm:1.5.0
│  └─ monaco-editor@npm:0.52.0 (via npm:^0.52.0)
│
└─ @hawtio/react@npm:1.5.0 [1e960]
   └─ monaco-editor@npm:0.52.0 (via npm:^0.52.0)
```
Maybe we need to update Hawtio standalone as well... Hawtio 4.2.0 should be broken.

grgrzybek:
Hmm - does this mean it'll work when monaco will be loaded from CDN?
What about offline/dmz environments with limited access to Internet?
Won't it work when loading from Hawtio origin?

docker/includes/security-headers-online.conf
mmelko: let me double-check this. But AFAIR  that loader from 0.43 was used also in patternfly and  0.5 monaco. But I have to double-check

mmelko:
@grgrzybek hmm valid point. This change comes from PF5 upgrade  and it should replace webpack requirements afaik

docker/includes/security-headers-online.conf
mmelko: @tadayosi ok, so patternfly-code-editor has dependency on the monaco-editor/react : https://github.com/patternfly/patternfly-react/blob/a3ffb39a0cc2c9130f7db86ad3186787ba12648e/packages/react-code-editor/package.json#L33 which has dependency on the monaco-editor/loader: https://github.com/suren-atoyan/monaco-react/blob/f7cac39fbad0f062dc66458831aaf57a7126dd40/package.json#L62 which contains: https://github.com/suren-atoyan/monaco-loader/blob/v1.4.0/src/config/index.js link to the version: `https://cdn.jsdelivr.net/npm/monaco-editor@0.43.0/min/` 

so the link is correct. @grgrzybek  Only question remains if we can replace this with "local" script. This is actually valid for both online and standalone.

tadayosi (approved):

grgrzybek:
We need to apply [these instructions](https://www.npmjs.com/package/@patternfly/react-code-editor#user-content-to-use-monaco-editor-as-an-npm-package-and-avoid-using-cdn)

grgrzybek:
Looks like it worked with [these instructions](https://www.npmjs.com/package/@patternfly/react-code-editor#to-use-monaco-editor-as-an-npm-package-and-avoid-using-cdn):
image

grgrzybek:
See hawtio/hawtio-next#1186

grgrzybek:
After hawtio/hawtio-next#1187 is merged and new `@hawtio/react` version is used in Hawtio online, please remove CDN host from CSP header. I'll do the same for hawtio/hawtio's ContentSecurityPolicyFilter.

Issue #583, opened 2024-12-13
Create Helm chart for K8s installation


eshepelyuk:
Currently proposed installation methods via CLI are not convenient. It would be great improvement to have  a Helm chart to be able to install Hawtio online in uniform way with another K8s applications.

phantomjinx:
Hey @eshepelyuk 

We could look at that, yes. If you are able to construct a helm chart for us, that would be awesome!

In the meantime, have you considered installing via [OperatorHub](https://operatorhub.io/operator/hawtio-operator)?

eshepelyuk:
> We could look at that, yes. If you are able to construct a helm chart for us, that would be awesome!

Hmm. not sure what do you mean, with 
> construct a helm chart for us

You already have all those manifests, the only difficulty is to publish it. And this is defined by your oganization, external contributors can't handle this.

> In the meantime, have you considered installing via [OperatorHub](https://operatorhub.io/operator/hawtio-operator)?

OperatorHub is smth that requires OLM installation? If yes, this is also not an option.

phantomjinx:
https://github.com/hawtio/hawtio-online/pull/636

PR #636, opened 2025-03-17
Implements helm charts for installing #583

phantomjinx:


tadayosi (approved):
Not so familiar with Helm either, but it looks good to me.

jsolovjo (commented):
Looks good to me:
- The instructions are clear;
- HawtIO Online got deployed smoothly with no issues;
- I also tried to deploy HawtIO with `---set mode=cluster` to see whether the property will get applied or not, and it worked;
- The only small suggestion for improvements is to replace `helm show values hawtio-online` by `helm show values hawtio/hawtio-online` in `docs/install/helm-charts.md`. Otherwise, it will throw `Error: non-absolute URLs should be in form of repo_name/path_to_chart, got: hawtio-online` error;

PR #673, opened 2025-05-13
Helm chart: add service and ingress annotations and service type

john9x:
Hi!

I propose to add customization for service and ingress annotations and service type.


phantomjinx (approved):
LGTM

PR #801, opened 2025-12-05
Push gateway image, latest tag

john9x:
Hi!

I propose to fix github action: push gateway image, fix `latest` tag


phantomjinx (approved):
Yeah. Looks good.

PR #947, opened 2026-07-27
HAWNG-2228: Corrects the real_ip_from settings in the nginx config

phantomjinx:
* nginx.sh
  * Instead of assuming a standard ip address class and defaulting to an all-address default, simply pass the 24 subnet mask of the actual IP address through to the nginx comfig file.

* nginx-gateway.conf.template
  * Set multiple real_ip_from properties for the standard ip address classes and the custom subnet injected from the runtime script
  * This allows for the ip address subnet to be trusted and so where applicable (rate limiting) use the client ip address rather than the server address.
  * Avoiding the 0.0.0.0/0 subnet is the crucial aspect as any endpoints that expect to use 127.0.0.1 (localhost) will be respected rather than substituting in the client ip address.

* patches ...
  * Provides a patch for providing a custom nginx configuration if required by users

See [HAWNG-2228](https://redhat.atlassian.net/browse/HAWNG-2228) for more information.

### DIFF
--- a/deploy/patches/override-nginx-config/hawtio-nginx-override-configmap.yml
+++ b/deploy/patches/override-nginx-config/hawtio-nginx-override-configmap.yml
@@ -1,0 +1,269 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: hawtio-nginx-template-override
+data:
+  nginx-gateway.conf.template: |
+    # vim: set filetype=nginx:
+
+    error_log /dev/stderr ${NGINX_LOG_LEVEL};
+
+    # IP Resolution
+    real_ip_header X-Forwarded-For;
+
+    #
+    # Overridden hardcoded variable
+    # Set to the class of the IP address of the
+    # hawtio-online pod / nginx server, eg.
+    # if ip is identified as 10.217.0.137
+    # then set to 10.0.0.0/8
+    #
+    set_real_ip_from 10.0.0.0/8;
+
+    # Define Zones
+
+    # Zone A: Sensitive APIs (Login, Management)
+    # Strict: 10 reqs/sec. Good for brute-force protection.
+    limit_req_zone $binary_remote_addr zone=api_limit:10m rate=10r/s;
+
+    # Zone B: Static Assets (Files)
+    # Fast: 50 reqs/sec.
+    limit_req_zone $binary_remote_addr zone=static_limit:10m rate=50r/s;
+
+    # Zone C: Kubernetes Proxy (High Traffic) - NEW
+    # We allow a high rate (30r/s) because the dashboard is chatty.
+    limit_req_zone $binary_remote_addr zone=proxy_limit:10m rate=30r/s;
+
+    # Zone D: Connections
+    limit_conn_zone $binary_remote_addr zone=addr_conn:10m;
+
+    proxy_cache_path /var/cache/nginx/pods levels=1:2 keys_zone=pods:2m max_size=4m inactive=60m use_temp_path=off;
+    proxy_cache_path /var/cache/nginx/rbac levels=1:2 keys_zone=rbac:256k max_size=3m inactive=60m use_temp_path=off;
+    proxy_cache_path /var/cache/nginx/rbac2 levels=1:2 keys_zone=rbac2:256k max_size=3m inactive=60m use_temp_path=off;
+
+    map $uri $new {
+        / /online/;
+        /online /online/;
+    }
+
+    map $http_upgrade $connection_upgrade {
+        default upgrade;
+        ''      close;
+    }
+
+    server {
+        listen              ${LISTEN_SERVER_PORT};
+        server_name         localhost;
+        ${SERVING_SSL_CERTIFICATE}
+        ${SERVING_SSL_KEY}
+        absolute_redirect   off;
+        gzip                on;
+        gzip_types text/plain application/javascript text/javascript;
+        root                /usr/share/nginx/html/;
+
+        # Keep the tunnel open for 1 hour even if silent
+        proxy_read_timeout 3600s;
+        proxy_send_timeout 3600s;
+        # Limit the keepalive of connections (60s default)
+        keepalive_timeout 60s;
+
+        # Do not show server version
+        server_tokens off;
+
+        # Performance tuning
+        subrequest_output_buffer_size ${NGINX_SUBREQUEST_OUTPUT_BUFFER_SIZE};
+        client_body_buffer_size       ${NGINX_CLIENT_BODY_BUFFER_SIZE};
+        proxy_buffers                 ${NGINX_PROXY_BUFFERS};
+
+        # For debugging location rewrite
+        #rewrite_log on;
+
+        if ($request_method !~ ^(GET|HEAD|POST)$ ) {
+          return 444;
+        }
+
+        include /etc/nginx/includes/security-headers-common.conf;
+
+        # Only accept newer ssl protocols (if applicable)
+        ${SERVING_SSL_PROTOCOLS}
+
+        if ($new) {
+            rewrite ^ $new redirect;
+        }
+
+        error_page 403 /hawtio-403.html;
+        location = /hawtio-403.html {
+            internal;
+        }
+
+        error_page 404 /hawtio-404.html;
+        location = /hawtio-404.html {
+            internal;
+        }
+
+        error_page 500 502 503 504 /hawtio-50x.html;
+        location = /hawtio-50x.html {
+            internal;
+        }
+
+        location /auth/logout {
+            limit_req zone=api_limit burst=20 nodelay;
+
+            if ( $arg_redirect_uri = '') {
+              return 200 "Acknowledge logout but nothing further to do";
+            }
+
+            include /etc/nginx/includes/clear-storage-headers.conf;
+            include /etc/nginx/includes/security-headers-common.conf;
+
+            proxy_pass                    ${HAWTIO_ONLINE_GATEWAY_APP_PROTOCOL}://localhost:${HAWTIO_ONLINE_GATEWAY_APP_PORT}/logout;
+            proxy_pass_request_headers    on;
+            proxy_pass_request_body       on;
+            proxy_redirect                off;
+            proxy_ssl_verify              off;
+            proxy_ssl_trusted_certificate /var/run/secrets/kubernetes.io/serviceaccount/ca.crt;
+            proxy_ssl_session_reuse       on;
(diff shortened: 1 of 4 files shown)


docker/nginx.sh
```diff
@@ -47,25 +47,14 @@ generate_nginx_gateway_conf() {
...
-     export REAL_IP_FROM="192.168.0.0/16"
-  else
-     # Fallback: If we can't determine the private net, we must default to 0.0.0.0/0
-     # to ensure the app works, though it reduces the precision of the rate limiting.
-     echo "WARNING: Could not determine private subnet from IP ${LOCAL_IP}. Defaulting 'set_real_ip_from' to 0.0.0.0/0"
-     export REAL_IP_FROM="0.0.0.0/0"
-  fi
+  # Calculate the /24 subnet for this specific pod
+  # Example: 100.96.1.5 → 100.96.1.0/24
+  CUSTOM_POD_NETWORK=$(echo "${LOCAL_IP}" | awk -F. '{print $1"."$2"."$3".0/24"}')
```
grgrzybek: No need to handle IP6?

deploy/patches/override-nginx-config/hawtio-nginx-override-configmap.yml
```diff
@@ -0,0 +1,269 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: hawtio-nginx-template-override
+data:
+  nginx-gateway.conf.template: |
+    # vim: set filetype=nginx:
```
grgrzybek, line 7: Am I right that this fragment is in yaml to be able to patch the real `nginx-gateway.conf.template` at deployment time? Why not change the file without passing it through yaml?

deploy/patches/override-nginx-config/hawtio-nginx-override-configmap.yml
```diff
@@ -0,0 +1,269 @@
...
+    real_ip_header X-Forwarded-For;
+
+    #
+    # Overridden hardcoded variable
+    # Set to the class of the IP address of the
+    # hawtio-online pod / nginx server, eg.
+    # if ip is identified as 10.217.0.137
+    # then set to 10.0.0.0/8
+    #
+    set_real_ip_from 10.0.0.0/8;
```
grgrzybek, line 21: Just to confirm my limited knowledge about this.
Does this config assume there's _another_ reverse proxy (ingress? some openshift/kubernetes nginx load-balancer?) that actually uses:
```
proxy_set_header X-Forwarded-For   $remote_addr;
```
?

and then is `10.0.0.0/8` a range related to podIP/clusterIP? What if the cluster network:
 - uses IP6?
 - uses different network?

grgrzybek (approved):

docker/nginx-gateway.conf.template
```diff
@@ -4,7 +4,14 @@ error_log /dev/stderr ${NGINX_LOG_LEVEL};
 
 # IP Resolution
 real_ip_header X-Forwarded-For;
-set_real_ip_from ${REAL_IP_FROM};
+
+# Trust standard private networks
+set_real_ip_from 10.0.0.0/8;
+set_real_ip_from 172.16.0.0/12;
+set_real_ip_from 192.168.0.0/16;
```
tadayosi: just wondering if there's no security implication with this allowance.

tadayosi (approved):

docker/nginx.sh
phantomjinx: Interesting question. I don't think any OpenShift / Kube cluster has implemented IPv6 in their containers. I shall investigate.

deploy/patches/override-nginx-config/hawtio-nginx-override-configmap.yml
phantomjinx: This is a replacement for the entire file. The resource (configmap) is yaml while the data is text file. The patch is not actually used during deployment but I wanted to keep it in the codebase since it was useful for the workaround to the original problem.

docker/nginx-gateway.conf.template
grgrzybek: From my perspective, this is fine, but many CVE descriptions mention recently "malicious, authenticated, insider attackers"...

deploy/patches/override-nginx-config/hawtio-nginx-override-configmap.yml
phantomjinx: Then the ip address in the X-Forwarded-For will not be trusted and the pod ip address will be used instead for all nginx endpoints. This only now has implications for the rate bursting limits in stopping DDOS attacks.

docker/nginx-gateway.conf.template
phantomjinx: ~~So these denote likely custom IP address subnets that would be trusted by default by nginx and if the X-Forward-For header uses a client ip address within these ip ranges, that address would be substituted in any endpoints that nginx serves. Given the number of endpoints has been reduced anyway due to making the gateway now do all the work of negotiating with the jolokia apps, the security implications should be minimal in my view.~~

This seems to be a fundamental misunderstanding on my part so re-investigating the security implications now.

deploy/patches/override-nginx-config/hawtio-nginx-override-configmap.yml
grgrzybek: ok. So this setup is: trust that when $remote_addr is from 10.0.0.0/8 then the _actual_ remote address comes from `X-Forwarded-For` header.
Which means some other proxy set `X-Forwarded-For` correctly (instead of an attacker with `curl -H'X-Forwarded-For: anything-trusted'`?

phantomjinx: Am reviewing this now and it looks like a misunderstanding on my part. Will probably need to modify this to tighten up security implications.

docker/nginx-gateway.conf.template
grgrzybek: from my experiments with Fetch Metadata and jolokia/jolokia#621 `set_real_ip_from` should specify 100% trusted other reverse proxies that _create_ (instead of just _pass_) `X-Forwarded-For` (or `X-Real-IP`).
I spent some time talking to an agent and you have to assume there's a list of:
`client -> proxy1 -> proxy2 -> ... -> proxyN -> your server`. And you have to trust at least one proxy when looking up the chain and the last trusted SHOULD use:
```
proxy_set_header X-Forwarded-For $remote_addr;
```
and NOT:
```
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
```

so the actual "remote address" you get through the list of proxies is at least the first (looking from server to client) untrusted proxy's address (because you will not get really-real client address).

This is more of a documentation thing and assumption, not strict mechanism.

phantomjinx: That is what I have also determined and the PR now addresses this my adding `real_ip_recursive on` to the nginx config.

phantomjinx:
So with a further look at this and an engaging discussion with AI, I have distilled the requirements and come up with an appopriate solution:

### **Requirements**

1. **Zero-Touch Dynamic Autoconfiguration**
Automatically determine the trusted pod/router network CIDR at container startup without requiring manual Operator updates.
2. **Runtime Escape Hatch**
Allow operators in non-standard environments (e.g., Carrier-Grade NAT `100.64.0.0/10`, public IP pod networks, or multi-tenant proxies) to override the trusted network CIDR dynamically via an environment variable (`HAWTIO_TRUSTED_NETWORK_SUBNET`) without needing to rebuild the image.
3. **Multi-Stack Networking (IPv4 / IPv6)**
Dynamically adjust trusted CIDR boundaries depending on the address family:
  - **IPv4:** Mask to a `/16` boundary (`X.Y.0.0/16`). This ensures OpenShift Ingress/Router pods running on Node A and Hawtio pods running on Node B share the trusted rule across cross-node CNI overlays.
  - **IPv6:** Mask to a `/64` boundary (`X:Y:Z::/64`), the standard universal IPv6 subnet boundary.
4. **Sidecar & Loopback Isolation (No 403 Cascades)**
Strictly exclude loopback (`127.0.0.0/8`) from `set_real_ip_from`. This guarantees internal proxying between container and sidecar retains its internal identity, preventing subrequest access-control failures.
5. **Accurate Rate Limiting & DoS Protection**
Ensure `$binary_remote_addr` resolves to the genuine external client IP. This prevents legitimate users from being grouped into a single shared Router IP bucket (which causes false-positive `503 Service Unavailable` errors under heavy console usage) while ensuring rate limits (`api_limit`, `static_limit`, `proxy_limit`) effectively limit malicious individual callers.
6. **Multi-Hop Proxy Traversal**
Enable `real_ip_recursive on;` to safely handle multi-tier proxy chains (e.g., Cloudflare / AWS ALB $\rightarrow$ OpenShift Router $\rightarrow$ Hawtio Nginx).

### **Summary of Changes**

#### 1. `nginx.sh` Entrypoint Logic
- Checked for `HAWTIO_TRUSTED_NETWORK_SUBNET` first to allow runtime overrides.
- Evaluates address format:
  - IPv4 address $\rightarrow$ Calculates `${OCTET1}.${OCTET2}.0.0/16`.
  - IPv6 address $\rightarrow$ Calculates `${NIBBLE1}:${NIBBLE2}:${NIBBLE3}::/64`.
- Exported `${CUSTOM_POD_NETWORK}` into `envsubst`.

#### 2. `nginx-gateway.conf.template`

```nginx
set_real_ip_from ${CUSTOM_POD_NETWORK};

```

### **Security & Operational Impact Analysis**

| Area | After |
| --- | --- |
| Rate-Limiting Behavior | Client IPs are uniquely resolved into $binary_remote_addr. Each user receives their own rate-limit allocation. |
| Intra-Cluster Attack Surface | Scoped tightly to local pod SDN /16 or /64 segments or explicit operator-defined CIDRs. |
| Sidecar Health | Localhost (127.0.0.1) is never trusted for header substitution; internal sidecar traffic remains isolated. |
| Operator Overhead | Zero-touch auto-detection with an optional ENV escape hatch. |

---

tadayosi (approved):

docker/nginx-gateway.conf.template
grgrzybek: I didn't check this option, but looks good!
But I'm just wondering - what if the actual client (not a reverse-proxy in the chain) is also from the same subnet as one of the proxies? Will nginx just use the first value if _all_ addresses from the list in the header indicated by `real_ip_header` are matched by `set_real_ip_from` (i.e., are trusted?)

